PR #6, opened 2017-02-05
FIxed integer division to work in Python3

numshub:
Patched by 2to3 and fixed a minor error.
I tested and it works in Python3 now.


ndbroadbent:
Ah cool, thanks!

PR #32, opened 2017-09-05
Add Dockerfile to run as a Docker container

eana:
Hello there!

I hope PRs are welcomed :)

This is a great script! I think it would be cool if you could run it within a container. Therefore I decided to make this PR to be able to run your script as a Docker container.

As an example, you could run `download_photos.py` as a container by simply:

```
docker run -it --rm --name icloud -v $(pwd)/Photos:/data icloud_photos_downloader ./download_photos.py \
    --username=testuser@example.com \
    --password=pass1234 \
    --size=original \
    --recent 500 \
    --auto-delete \
    /data
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,3 +97,25 @@ This process can take around 5-10 minutes, so please wait a few minutes, then tr
 ```
 0 */3 * * * /path/to/icloud_photos_downloader/cron_script.sh
 ```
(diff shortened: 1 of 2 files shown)


ndbroadbent:
Awesome, thanks very much!

PR #36, opened 2017-09-25
Pyicloud updated their code to be 2sa instead of 2fa. Fixing on this end as well

dillanmills:


ndbroadbent:
Ah, great. Could you please add a `callable(requires_2sa)` line to provide backwards compatibility with previous versions of `pyicloud`? Thanks!

dillanmills:
okay, it's pushed.

ndbroadbent:
Whoops, I gave you the wrong suggestion. We needed to use `hasattr` instead of `callable`. I pushed the fix: 43026e6b3f18d75b3b628ad58ff3190df59d05fa

Thanks for letting me know about that change!

PR #51, opened 2017-12-23
Catching API errors produced by pyicloud 

jorgeas80:
I had a 503 unavailable error that crashed the process

### DIFF
--- a/download_photos.py
+++ b/download_photos.py
@@ -7,6 +7,8 @@ import socket
 import requests
 import time
 import itertools
+
+from pyicloud.exceptions import PyiCloudAPIResponseError
 from tqdm import tqdm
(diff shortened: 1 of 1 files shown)


ndbroadbent:
Thanks very much, and sorry for the delay! I've rebased and merged your commits

fnordahl:
@ndbroadbent @jorgeas80 The merging of this PR broke the re-authentication fix in #60. So either we must remove catching of PyiCloudAPIResponseError from download_photo function or move the fix in download() function to download_photo function.

ndbroadbent:
Oh, right!  Maybe we should also test the error message to make sure it matches `Invalid global session`.

PR #55, opened 2018-01-23
Avoid shadowing built-in file() function

svisser:
I noticed this because of Github's syntax highlighting of that part of the code.

This avoids the use of `file` as variable as this clashes with the built-in `file()` function in Python 2.7.


ndbroadbent:
Thanks!

PR #59, opened 2018-03-18
Add support for prompting for password

### LINKED ISSUE #58: No support for prompting for password
Giving password as a command line argument has some security risks as it will leave the password in the shell history and it is visible to other users on a system through the process list.

You can currently add password to PyIcloud's keychain but it would also be nice to have the download_photos command prompt for password when not provided.

fnordahl:
Fixes #58


ndbroadbent:
Thanks!

PR #63, opened 2018-03-29
Add .heic photo support

gneitzke:
iOS 11 added support for .heic (https://www.macworld.com/article/3226490/ios/ios-11-hevc-heif-and-what-you-need-to-know-about-these-new-video-and-photo-formats.html)


ndbroadbent:
Great, thanks!

PR #76, opened 2018-06-12
Avoid to download in the name of the file without the size is found

avouspierre:
This pull request avoid to download the cloud file if the name of the file is found in the local directory but without the extension of the size. This use case happens when you downloaded the photo directly with photos app (export original files). 
If the --size parameter is Original,  test if the NAME-original.JPG **or NAME.JPG** is found in the local directory, not download the file.


ndbroadbent:
Great, thanks! This sounds likes a good idea to me

Issue #103, opened 2018-09-26
Support SMTP-server without username / password


tkallevik:
Hi

Let me just start by thanking for this great project. I have been using it for a couple of months now, and it is working like a charm.

But back to the issue at hand. I have a local SMTP-server (Postfix) running on the server that execute the icloudpd, and handles some local deliveries. This SMTP-server is restricted to the local network, and do not have any authentication. What would be nice was if I could have smtp-username and smtp-password empty, but still get two-step authentication notifications.

Would it be possible for you to change the logic of "should_send_2sa_notification" to be if either the smtp-username or notification-email is not empty? And add an if-block around the smtp.login(smtp_email, smtp_password), making it only log in if either smtp-username or smtp-password is set?

If you don't have time, or don't want to prioritize this at this point, I can give it a try to change this myself and create a PR, but I am not a very well-versed python developer :)

ndbroadbent:
Hi @tkallevik, sure that sounds like a good change to me! Unfortunately I'm very busy at the moment, so it would be really helpful if you would be able to try making the change and sending a PR. But let me know if you get stuck, and I can take a look

tkallevik:
@ndbroadbent I have tried creating a PR now with the changes. I think it should be ok, but there is one "quirk" I know of.

You used the smtp_username as from address for the email being sent. So when this is not provided ,due to a SMTP-server without authentication, we have no address to use as a from address. To handle this I just used the notification-email. That means that for SMTP-servers without authentication the from address and to address will always be the same. But my thinking was that this was already the default case, if no notification email is specified, since then you use the smtp-username as the to address as well.

So I hope you agree with me, and think this is ok :)

ndbroadbent:
Yes that makes sense to me! And if that works for you, then that's fine. If someone needs to set the from_address as a command-line argument, then I can just add that later, but no-one has requested it yet.

So yes this is a good default

tkallevik:
Great, that was my thinking as well!

tkallevik:
@ndbroadbent Thank you so much for answering quickly, and following this up as fast as you have done. That is awesome :)

Will these changes be available through pip immediately, or do I have to wait for some release?

ndbroadbent:
There is sometimes a short delay but it's available now. Try: `pip install icloudpd==1.3.2`

tkallevik:
I just saw the comment on the PR that you had pushed the new version :) It has now been installed, and is working like a charm.

[...]

PR #104, opened 2018-09-26
#103 Add support for SMTP-server without authentication

tkallevik:
Pull request for issue #103

### DIFF
--- a/icloudpd/email_notifications.py
+++ b/icloudpd/email_notifications.py
@@ -12,6 +12,7 @@ def send_2sa_notification(
 ):
     """Send an email notification when 2SA is expired"""
     to_addr = to_addr if to_addr else smtp_email
+    from_addr = smtp_email if smtp_email else to_addr
     logger = setup_logger()
     logger.info("Sending 'two-step expired' notification via email...")
     smtp = smtplib.SMTP()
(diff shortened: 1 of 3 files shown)


icloudpd/email_notifications.py
```diff
@@ -12,14 +12,17 @@ def send_2sa_notification(
...
     logger = setup_logger()
     logger.info("Sending 'two-step expired' notification via email...")
     smtp = smtplib.SMTP()
     smtp.set_debuglevel(0)
     smtp.connect(smtp_host, smtp_port)
     if not smtp_no_tls:
         smtp.starttls()
-    smtp.login(smtp_email, smtp_password)
+
+    if smtp_email is not None or smtp_password is not None:
```
ndbroadbent: I think this should be `and` instead of `or`?

tkallevik: Maybe. It was a conscious choice on my part, because I was thinking that maybe you had an SMTP-server that required a username but the password was blank. Or maybe an SMTP-server without any username requirement, and only a password requirement.

But I am happy to change it to `and`, if you feel that it makes more sense.

ndbroadbent: Oh right, I just wasn't sure if those cases were possible with SMTP. But yes the `or` makes sense, we may as well support those cases as well!

ndbroadbent:
Awesome, thanks for this! I've merged the PR and pushed up version `1.3.2`.

PR #110, opened 2018-11-02
Multi-threaded downloads and basic caching

thechild:
I added a ThreadPoolExecutor to speed up the downloading process - seems to help a lot.  I also added some basic caching to speed up restarting a process that failed partially through.  I have 40k photos in my library that I'm trying to back up, so this has made a big difference so far.

Three issues with this approach:
1) I didn't update the progress bar at all.  It seems to still work fine though, but is kind of random in which file it shows at any given time.
2) Ctrl-C doesn't always work due to all the threads. I tried to fix this a little but if you just hit it a couple times it eventually works so I gave up.
3) Caching only sort of helps, since we still need to hit icloud to get the filename of each file to compare with the cache.  I tried on id as well, but that also requires a network hop so it doesn't speed up as much as possible.  Diving into the pyicloud internals more might show a better thing to cache on but I didn't find it quickly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ __pycache__
 .coverage
 htmlcov
 downloaded_photos_cache.p
+downloaded_ids_cache.p
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ __pycache__
 .idea
 .coverage
 htmlcov
+downloaded_photos_cache.p
--- a/icloudpd/autodelete.py
+++ b/icloudpd/autodelete.py
@@ -6,7 +6,7 @@ from icloudpd.logger import setup_logger
 from icloudpd.paths import local_download_path
 
 
-def autodelete_photos(icloud, folder_structure, directory):
+def autodelete_photos(icloud, folder_structure, directory, downloaded_photos):
     """
     Scans the "Recently Deleted" folder and deletes any matching files
     from the download directory.
@@ -27,3 +27,4 @@ def autodelete_photos(icloud, folder_structure, directory):
             if os.path.exists(path):
                 logger.info("Deleting %s!", path)
                 os.remove(path)
+                downloaded_photos.remove(path)  # remove from cache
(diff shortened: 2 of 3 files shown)


ndbroadbent:
Great, thanks! I agree that multi-threading is a good idea and will speed things up.

I'm not sure about the need for caching, because we can just check to see if the file exists on the disk, and like you say the network operations are a bottleneck. But pyicloud is doing paginated requests to return the first x photos, so a better way would be to skip any pages that have all been downloaded. We might need to add a new hook in pyicloud so we can see the current page, because at the moment all that pagination logic is hidden.

I think it's probably fine to only support Python > 3.2, but it might be better if we can also support 2.7 and maybe use a library, or some of the other concurrency modules. Oh and also I think the progress bar is fine, as long as it isn't totally broken. But I'm thinking it might be cool to show different progress bars for each thread, and I think tqdm supports that.

otherjoel:
Just to add my $0.02, caching would be nice if it allows me to move the downloaded files to another location in the future without making icloudpd "lose its place" in the photo stream or re-download them later on.

ianmagnone:
I pulled this new version to test and it definitely downloads faster. Easily 3-5x faster. I have 60k+ items so that makes a big difference.

The progress bar is very inaccurate though. It seems to jump through all the items then stops when it gets to 100%. Meanwhile it's continuing to download in the background.

I'm not sure the best way to present the current status while keeping the download speeds fast.

If checksum validation can be added in with these features this is a nearly perfect tool for me to keep an offline backup of my invaluable family photos and videos.

Thank you for all your efforts on this!

nurikk:
Got error python 3.4

`  File "/home/orangepi/icloud_photos_downloader/icloudpd/base.py", line 490, in main
    with ThreadPoolExecutor() as executor:
TypeError: __init__() missing 1 required positional argument: 'max_workers'`

ndbroadbent:
Merged in #132

PR #134, opened 2019-04-17
new commandline parameters fixes #106

### LINKED ISSUE #106: Download specific album
First, super thanks for an awesome tool. I'm using it and it's amazing!

I would like to have the option to download a specific album by his name. For example,  I want to download all my favorite photos.

Thanks in advance.
Ran

menkej:
I added two new parameters:

--list-albums (list all albums)
-a, --album (select the album to be downloaded)

With this it is possible to download a specific album only. I use this for my screensaver album. "All Photos" is still the default. I also made "directory" an option, so that it is not mandatory any more. I did not add a default target directory but I guess "." can be a valid default. What do you think about this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,15 @@ pip install icloudpd
 
     $ icloudpd <download_directory>
                --username <username>
-               [--password <password>]
+               [-p, --password <password>]
+               [-d, --directory <directory>]
                [--cookie-directory </cookie/directory>]
                [--size (original|medium|thumb)]
                [--live-photo-size (original|medium|thumb)]
                [--recent <integer>]
                [--until-found <integer>]
+               [-a, --album <album>]
+               [-l, --list-albums]
                [--skip-videos]
                [--skip-live-photos]
                [--force-size]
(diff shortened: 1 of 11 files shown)


menkej:
Ok, maybe I should also have a look at the tests...

menkej:
> Ok, maybe I should also have a look at the tests...

menkej:
It seems like the user used for running the test-cases is active. Is it so, @ndbroadbent ? This influences the creation of test cases...

ndbroadbent:
> It seems like the user used for running the test-cases is active. Is it so, @ndbroadbent ? This influences the creation of test cases...

Hi, yes I created the test cases by using my own iCloud account, but then I recorded all of the API responses with VCR, so they can be run offline, on CI, etc.

It might be tricky to add new test cases for VCR, but you could do the same thing by setting up a VCR cassette and entering your own iCloud password, and them making sure to remove the password from the recorded responses.

ndbroadbent:
Hi @menkej, thanks very much for all your work! I've fixed the remaining issue, will merge your changes into master, and release this under version: 1.4.4. Thanks!

PR #140, opened 2019-06-21
comment out section that causes error 'PhotoAsset object has no attribute called "item_type"'

danielfmolnar:


ndbroadbent:
Hello, thanks for the PR! Unfortunately I am not able to accept this because it does break some functionality, and causes some tests to fail. But I might be able to dive in fix the underlying cause at some point. Thanks though!

PR #149, opened 2020-05-17
Fix so that the logger only returns one global logger, allowing logging

koldinger:
settings to be used globally, and eliminating the overlogging when a log-level is specified.

### DIFF
--- a/icloudpd/logger.py
+++ b/icloudpd/logger.py
@@ -32,21 +32,27 @@ class IPDLogger(logging.Logger):
             self.tqdm.write(message)
 
 
+logger = None
 def setup_logger(loglevel=DEBUG):
     """Set up logger and add stdout handler"""
-    logging.setLoggerClass(IPDLogger)
-    logger = logging.getLogger("icloudpd")
-    logger.setLevel(loglevel)
-    has_stdout_handler = False
-    for handler in logger.handlers:
-        if handler.name == "stdoutLogger":
-            has_stdout_handler = True
-    if not has_stdout_handler:
-        formatter = logging.Formatter(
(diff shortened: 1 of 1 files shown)


MrBenFTW:
How did you get it working? I have an issue with albums, #150

koldinger:
Can't imagine that this is related.   This merely uses a common logging object, so that the settings get reflected everywhere, rather than a separate logging object for each component.

MrBenFTW:
I understand, just asking how you got your build to work, since the master is broken

koldinger:
I just ran it.   Seems to be fine.

ndbroadbent:
Hi @koldinger, thanks very much for the PR, and sorry for the delay! I was just wondering if you could please take a look at the pylint errors in the Travis CI build? (If you are still interested)

```
3.73s$ pylint icloudpd
************* Module icloudpd.logger
icloudpd/logger.py:35:0: C0103: Constant name "logger" doesn't conform to UPPER_CASE naming style (invalid-name)
icloudpd/logger.py:38:4: C0103: Constant name "logger" doesn't conform to UPPER_CASE naming style (invalid-name)
icloudpd/logger.py:38:4: W0603: Using the global statement (global-statement)
icloudpd/logger.py:39:4: R1705: Unnecessary "else" after "return" (no-else-return)
```

menkej:
I like using a singleton for logging, so I understand the intention. Anyways pylint has a good point not to use globals, etc. But to solve this we should restructure the program in a more object-oriented way (to be a class), so that we could just put the logger into an instance variable. But this is quite a big change...

menkej:
Thanks again for your pull request. We added a different solution with #194

PR #165, opened 2020-10-02
Handle icloud error for to many starts fixes #163

### LINKED ISSUE #163: pyicloud_ipd.exceptions.PyiCloudAPIResponseError: private db access disabled for this account.
hi.
i'm getting alot of those lately:
`pyicloud_ipd.exceptions.PyiCloudAPIResponseError: private db access disabled for this account.  Please wait a few minutes then try again.  The remote servers might be trying to throttle requests. (ACCESS_DENIED)
`
with my icloud account. my sons account never gets this error. 

i suppose thats because my photo library is much bigger. what is the actual limit of requests? is it per hour? is there any info of how this throttling mechanism works?

menkej:
* ICloud obviously has some kind of throtteling/prevention for DOS
* some cosmetical changes p.e. cleanup of python 2

### DIFF
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -28,6 +28,8 @@ from icloudpd import exif_datetime
 from icloudpd import constants
 from icloudpd.counter import Counter
 
+from pyicloud_ipd.exceptions import PyiCloudAPIResponseError
+
 try:
     import Queue as queue
 except ImportError:
@@ -133,7 +135,8 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 )
 @click.option(
     "--set-exif-datetime",
-    help="Write the DateTimeOriginal exif tag from file creation date, if it doesn't exist.",
+    help="Write the DateTimeOriginal exif tag from file creation date, " +
+    "if it doesn't exist.",
     is_flag=True,
 )
 @click.option(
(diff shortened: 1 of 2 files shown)


ndbroadbent:
Thanks! I was just wondering, what do you think about pausing for 10-20 minutes and retrying, instead of quitting? It might be nice to show a message saying "Apple has started throttling API requests, so we need to pause for 10 minutes before resuming downloads."
I think that might be a good workaround so that people can leave it running overnight to download large libraries

menkej:
> what do you think about pausing for 10-20 minutes and retrying, instead of quitting? It might be nice to show a message saying "Apple has started throttling API requests, so we need to pause for 10 minutes before resuming downloads."

This would only make sense for overnight batch runs. This happens right after start (when you have started the script too often before), so quitting here would be right IMHO.

For batch runs this would happen during downloads. This is not handled yet, as I could not reproduce it by now. That would be in another place in the source code. There it would make sense to just pause for a longer time and retry.

Currently there is a "hard retry" after 5 secs for x times during downloads. Maybe it would just help to wait 5 secs in the first loop, then 10, then 20, then 40, then 80 secs. What do you think?

But, as said, I have not seen this error during downloads yet.

ndbroadbent:
That's a good point, thanks! I forgot that this would usually crash right at the beginning of the script. And yes I think the exponential back-off for the retries would be a good idea.

ndbroadbent:
There are just some linting errors to look into, but otherwise it looks good! https://travis-ci.org/github/ndbroadbent/icloud_photos_downloader/builds/732226416

```
3.94s$ pylint icloudpd
************* Module icloudpd.base
icloudpd/base.py:282:2: W0511: TODO: come up with a nicer message to the user. For now take the exception text (fixme)
icloudpd/base.py:281:4: C0103: Variable name "e" doesn't conform to snake_case naming style (invalid-name)
icloudpd/base.py:284:8: R1722: Consider using sys.exit() (consider-using-sys-exit)
icloudpd/base.py:31:0: C0411: third party import "from pyicloud_ipd.exceptions import PyiCloudAPIResponseError" should be placed before "from icloudpd.logger import setup_logger" (wrong-import-order)
```

AndreyNikiforov (approved):
/legit

ndbroadbent:
Awesome, thank you!

Issue #175, opened 2020-10-19
Support for multiple languages.


wclebb:
Please allow me to use Google Translate to describe my needs.
Some time ago, I learned that icloudpd can help me back up iCloud photos to NAS, which is what I want.

Later, when executing the command icloudpd, errors were often reported;
such as
INFO     Great, you're all set up. The script can now be run without user interaction until 2SA expires.
You can set up email notifications for when the two-step authentication expires.
(Use --help to view information about SMTP options.)
Traceback (most recent call last):
  File "/home/wclebb/.local/bin/icloudpd", line 8, in <module>
    sys.exit(main())
  File "/home/wclebb/.local/lib/python3.8/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/wclebb/.local/lib/python3.8/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/wclebb/.local/lib/python3.8/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/wclebb/.local/lib/python3.8/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/wclebb/.local/lib/python3.8/site-packages/icloudpd/base.py", line 292, in main
    directory = os.path.normpath(directory)
  File "/usr/lib/python3.8/posixpath.py", line 336, in normpath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType

I consulted V2ex（https://www.v2ex.com/t/716199） and got answers from netizens. This is due to an error in Chinese compatibility. I need to try to delete all related Chinese albums in iCloud. After I tried to delete all Chinese albums, I finally executed it correctly and downloaded the first photo. This is good news for me having been lingering on this error for three days.

But the bad news is that icloudpd has an error due to Chinese compatibility.
Therefore, I hope to be able to obtain compatibility support for multiple languages, because if I use Chinese or other languages, there will definitely be such a demand;

Therefore, thanks for providing icloudpd.

AndreyNikiforov:
@wclebb I suspect the issue is related to environment and/or python setup.

I added test in #176 to validate and it works on my windows and in docker. To check your setup, you can 1) run `docker build .` 
 to validate that dockerized python works with your files and 2) `py.test --cov=icloudpd --timeout=3 --cov-report term-missing` 
[...]

PR #176, opened 2020-10-19
add Chinese base path test; refactor tests for multi platform support

AndreyNikiforov:
writing tests to confirm compatibility of the platform with Chinese chars for #175

### DIFF
--- a/icloudpd/autodelete.py
+++ b/icloudpd/autodelete.py
@@ -23,7 +23,7 @@ def autodelete_photos(icloud, folder_structure, directory):
         download_dir = os.path.join(directory, date_path)
 
         for size in [None, "original", "medium", "thumb"]:
-            path = local_download_path(media, size, download_dir)
+            path = os.path.normpath(local_download_path(media, size, download_dir))
             if os.path.exists(path):
                 logger.info("Deleting %s!", path)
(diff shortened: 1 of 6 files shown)


menkej (approved):
Great work and a long diff to scan over! ;-)
Didn't know about normpath-method. Makes totally sense!
I can't test this on Windows right now. I assume os.path.join() handles slash vs backlash  correctly... 

Thinking about the logger issues we're having (also with a pull request) I'm more and more thinking about putting this program into a more object oriented (classes based) structure. So we could use just one logger instance for all... Nothing really todo with this change, but on my mind...

AndreyNikiforov:
> I can't test this on Windows right now. I assume os.path.join() handles slash vs backlash correctly...

Yes, you are correct, join() handles proper slashes

ndbroadbent:
Thanks for all the great work so far @AndreyNikiforov and @menkej! Please ping me whenever you would like to release a new version and I'm happy to take a look.

I'd also be happy to hand this over to you guys if you wanted to manage some of the releases

AndreyNikiforov:
> I'd also be happy to hand this over to you guys if you wanted to manage some of the releases

Just created #181 to address/improve that situation

PR #177, opened 2020-10-20
add check for missing --directory parameters

AndreyNikiforov:
Checking that --directory parameter is specified. Related to #175

### DIFF
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -232,6 +232,7 @@ def main(
         threads_num,
 ):
     """Download all iCloud photos to a local directory"""
+
     logger = setup_logger()
     if only_print_filenames:
         logger.disabled = True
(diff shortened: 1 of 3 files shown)


menkej:
Good to check this. I started digging into Click (without success yet). There should be a possibility to let this be checked there. Roughly described its like "-d is always required, if not using --list-albums", right?

menkej (approved):
Fine for me. Better then without this change, even though I would prefer to have this handled via click (which should somehow be possible, and if not would be worth a feature request to Click)...

AndreyNikiforov:
Click custom class looked as unnecessary complexity for our case. E.g. https://stackoverflow.com/questions/44247099/click-command-line-interfaces-make-options-required-if-other-optional-option-is I did not find simpler implementation using Click. If found, we can switch implementation (later).

Issue #155, opened 2020-06-16
IOError while writing to file. You might have run out of disk space, or file might be too large for your OS


HarrisonOates:
I've run the script through once, and it worked alright, however it did not download all of the photos. I am getting the error 
`OError while writing file to #### You might have run out of disk space, or the file might be too large for your OS. Skipping this file...`
I'm on an Arch Linux machine.

nihelmasell:
I have the same issue.

Monsterbacke007:
I have the same issue: OS Win10 Pro trying to download onto a network drive (NAS).

AndreyNikiforov:
@LoonyGryphon @Monsterbacke007 assuming that disk has enough space for the file (and destination system supports large files), do you mind trying the process again? Since it was skipped during the error, next run will try to download it again and we'll see if the problem is specific to the file or intermittent.

nihelmasell:
I used this solution https://github.com/ndbroadbent/icloud_photos_downloader/issues/150#issuecomment-699099138 because I also had the issue mentioned in that thread. After doing that, I didn't have this problem again (it "magically" solved two issues at once).

Monsterbacke007:
@nihelmasell I am getting a syntax error trying this on Win10 Pro: As I am new to Python can someone please help?

`@property
    def albums(self):
        if not self._albums:
            self._albums = {name: PhotoAlbum(self, name, **props)
                            for (name, props) in self.SMART_FOLDERS.items()}

            for folder in self._fetch_folders():
                # FIXME: Handle subfolders
                if folder['recordName'] in ('----Root-Folder----',
                    '----Project-Root-Folder----') or \
                    (folder['fields'].get('isDeleted') and
                     folder['fields']['isDeleted']['value']):
                    continue

                folder_id = folder['recordName']
                folder_obj_type = \
                    "CPLContainerRelationNotDeletedByAssetDate:%s" % folder_id
                folder_name = base64.b64decode(
            try:
                folder['fields']['albumNameEnc']['value']).decode('utf-8')
            except KeyError:
                continue
            
                query_filter = [{
                    "fieldName": "parentId",
                    "comparator": "EQUALS",
                    "fieldValue": {
                        "type": "STRING",
                        "value": folder_id
                    }
                }]

                album = PhotoAlbum(self, folder_name,
                                   'CPLContainerRelationLiveByAssetDate',
                                   folder_obj_type, 'ASCENDING', query_filter)
[...]

PR #178, opened 2020-10-21
set default threads to 1

AndreyNikiforov:
What: sets default `--threads-num` parameter to 1
Why: multi-threaded download causing issues when duplicate files exist in iCloud or on disk. Changing default will help reduce a number of issues. #155
Note: Discussion about deprecating multi-threaded downloading is separate from this fix


menkej (approved):
Actially I was thinking about the same. Change is simple, side effects not expected.

PR #182, opened 2020-10-22
Delete first and size check

Wayne2Wen:
The code should do the delete task first, then do the download task, and add the size check before deleting.

**Why:**
When you took a photo with your iPhone, then copied the photo for some reason, and then deleted the copied photo. 
At this time, the deleted photo has the same file name as the original photo. If the download task is performed before the deletion task, the original photo will be deleted by subsequent deletion tasks.

**TODO**
Fix tests run....

### DIFF
--- a/icloudpd/autodelete.py
+++ b/icloudpd/autodelete.py
@@ -17,15 +17,20 @@ def autodelete_photos(icloud, folder_structure, directory):
 
     recently_deleted = icloud.photos.albums["Recently Deleted"]
 
-    for media in recently_deleted:
-        created_date = media.created
-        date_path = folder_structure.format(created_date)
-        download_dir = os.path.join(directory, date_path)
+    all_photos_id = []
+    for photo in icloud.photos.all:
+        all_photos_id.append(photo.id)
 
-        for size in [None, "original", "medium", "thumb"]:
-            path = os.path.normpath(
-                local_download_path(
-                    media, size, download_dir))
-            if os.path.exists(path):
-                logger.info("Deleting %s!", path)
-                os.remove(path)
+    for media in recently_deleted:
+        if not (media.id in all_photos_id):
+            created_date = media.created
+            date_path = folder_structure.format(created_date)
+            download_dir = os.path.join(directory, date_path)
+
+            for size in [None, "original", "medium", "thumb"]:
+                path = os.path.normpath(
+                    local_download_path(
+                        media, size, download_dir))
+                if os.path.exists(path):
+                    logger.info("Deleting %s!", path)
+                    os.remove(path)


menkej (commented):
I need to think about the "why?" of this change. I did not have this issue yet. I'll try to reproduce.

The way you changed base.py the deletion is also triggered if you only --list-albums. This feels not right to me. @AndreyNikiforov  what do you think?

Wayne2Wen:
> I need to think about the "why?" of this change. I did not have this issue yet. I'll try to reproduce.
> 
> The way you changed base.py the deletion is also triggered if you only --list-albums. This feels not right to me.

Put the delete task in the front and the download task in the back, which will indeed trigger the deletion, but it will be downloaded later. But if the deletion task is after the download, the photo will be deleted eventually and there is no chance to download.

Wayne2Wen:
> --list-albums

Fixed the situation with only --list-albums. 

I only recently started learning python in order to use icloudpd. It’s definitely not very good. If you confirm the problem I mentioned above and have a better solution, of course it’s great.

AndreyNikiforov:
@Wayne2Wen I am trying to understand the use case:
1. Take a picture with iPhone (say, it is img_1234.jpg in iCloud)
1. Copy image in iCloud/on iPhone (now there are two files in iCloud with img_1234.jpg name with exactly the same content and same size, same metadata)
1. Use iCloudPD to download photos (will have one file img_1234.jpg, the second will be skipped as dup because name and size match)
1. Delete the copy of the picture in iCloud/iPhone (now we'll have only one img_1234.jpg is in the iCloud; it actually does not matter which one was deleted because they were identical)
1. Run iCloudPD again (download part will skip the file as dup, but delete part will delete it because file with the same name is in delete folder in iCloud)

If that is correct, then yes, we do have a problem with deletion.

Wayne2Wen:
@AndreyNikiforov Yes, it is the same in my deletion case.

If this bad case is fixed, the size check before deletion is not needed, I will revert it.

Wayne2Wen:
I also found that if the copied photo has been edited, the editing effect will be lost when downloaded. Just like the original photo, it would be better if the editing effect can be retained.

boredazfcuk:
> I also found that if the copied photo has been edited, the editing effect will be lost when downloaded. Just like the original photo, it would be better if the editing effect can be retained.

Photo edits are stored as .AAE files. They are just simple .xml files which detail the changes made. They aren't images in their own right.

When I look at iCloud.com, I'm only shown the edit. If you say that only the original file is downloaded, it would mean the edit file is also uploaded to iCloud and that the website creates the version you see on the web using both the files. Even if the .AAE file was available to download from iCloud, there would need to be a separate process to possibly apply the .AAE file to the original image to create the edited version.

AndreyNikiforov:
> @AndreyNikiforov Yes, it is the same in my deletion case.
> 
> If this bad case is fixed, the size check before deletion is not needed, I will revert it.

@Wayne2Wen can you please think of a way to codify the scenario in a test and then add your solution to prove that it fixes the issue?

Wayne2Wen:
@Wayne2Wen Yes, can you please provide me with a yml file that meets the above case？I am not good at editing the iCloud mock files

AndreyNikiforov:
> @AndreyNikiforov Yes, can you please provide me with a yml file that meets the above case？I am not good at editing the iCloud mock files.

Neither am I. I believe these files were recordings. @ndbroadbent do you have a recommendation on the best way to add another test case involving iCloud data?

ndbroadbent:
> @ndbroadbent do you have a recommendation on the best way to add another test case involving iCloud data?

These were all real test cases that I recorded using my own iCloud account, and then I scrubbed any sensitive information from the recordings and updated the authentication to use some fake keys. The idea is that you can run the tests in "real life" once to record all of the YML files using your real account, and then replay the sanitized recordings during the test suite (which doesn't require any real requests to Apple.)

icloudpd/base.py
```diff
@@ -367,6 +367,9 @@ def photos_exception_handler(ex, retries):
         photos_enumerator = tqdm(photos, **tqdm_kwargs)
         logger.set_tqdm(photos_enumerator)
 
+    if auto_delete:
```
AndreyNikiforov, line 370: According to the scenario https://github.com/icloud-photos-downloader/icloud_photos_downloader/pull/182#issuecomment-714911882 same file will be in `deleted` list and in `all albums`. If we delete it first, we'll have to re-download it later, correct? If correct, then it is not an ideal situation, but still provided correct state at the end.

AndreyNikiforov (changes requested):

icloudpd/base.py
Wayne2Wen: Ok.  I am working on this.

Wayne2Wen:
@AndreyNikiforov  Currently, it is too difficult for me to generate a suitable yml file for testing. The fix is ​​effective for me. I will use my own branch to synchronize photos for the time being.

Maybe after some time I will try again, you can turn off this pull request first.

AndreyNikiforov:
@Wayne2Wen Thanks for trying!

PR #186, opened 2020-10-25
Test on win action

menkej:
I updated the test actions and some documentation. Python 3.4 is no longer supported, so I updated the readme to go for 3.6

- initially I also wanted an automatic test to be run on Windows (that's the reason for the branch name). But this is a little more work... So I reverted to ubuntu only.
- I used separate test runs for the different python versions 3.6-3.9. This led to sometimes all tests running ok, sometimes one or two not working. Usually ending up with `FAILED tests/test_download_photos.py::DownloadPhotoTestCase::test_until_found` :-(
- then I switched to matrix mode. Really, really cool stuff! Testing multiple os/python-Version combos in just a hand full lines of code! :-)
- Consequence: if one job fails, the other jobs are aborted.

test_until_found is really annoying. Sometimes it works, sometimes it does not. Not at all related to these changes. I experience the same on my local Mac...


AndreyNikiforov (approved):

AndreyNikiforov:
> test_until_found is really annoying. Sometimes it works, sometimes it does not. Not at all related to these changes. I experience the same on my local Mac...

@menkej we can address test_until_found issue on its own as it is not related to this work. I agree it is annoying and we need to address it soon and it hurts our dev process.

Issue #188, opened 2020-10-26
Inconsistent result of test_until_found


AndreyNikiforov:
test_until_found behaves inconsistently. Need to understand the pattern and fix it.

See #186 as an example

menkej:
I have no Idea what is happening here. It works like a charm on my local Mac installation with python 3.8. The Actions on GitHub run on Linux. This should not really make a difference... And also different versions in the matrix test keep failing. Maybe we should revert to just one test-version again?

AndreyNikiforov:
> Maybe we should revert to just one test-version again?

If we claim that we support 3.6+ then we should test for these versions. We need to figure out why test is failing and fix it.

On Windows platform, for example, I suspected that creating large files takes a while and test fails with timeout.

menkej:
I have configured a 5 min limit for the tests right now. Do you think it is too low? AND: the Action on Github is not failing because of timeout, as the whole action fails in < 5 mins...

Funny enough I have (sometimes) other tests failing on my local installation:
```
test_download_photos_and_set_exif
test_download_photos_and_set_exif_exceptions
test_invalid_creation_year
```
Usually failing because of a timeout. They run for ages, if I would not use a timeout... Do the tests run ok on your installation? I think I'll set up a linux box tonight to run the tests in a different environment as well. Maybe I can find out something else...

I have no problem with using a newer python-version. Most current distros come with 3.8+ anyways today. This would also enable to use newer 3rd party libs. Even though python 3.6 is still receiving security patches till end of 2021 and python 3.7 till end of 2023... newer version (3.8/3.9) still receive also bugfixes, not only security... (taken from [here...](https://www.python.org/downloads/))

AndreyNikiforov:
I am using `docker buildx build .` and it was relatively stable. Tests on Windows are flaky. Increasing timeouts is a last resort IMO, we need to figure out what causing issues and fix that.

menkej:
finally the tests run smoothly on my Mac! Good job @AndreyNikiforov !

Issue #180, opened 2020-10-21
Remove multi-threaded downloads


AndreyNikiforov:
Multi-threaded downloading is causing issues with duplicate content #155 

Need to evaluate if a promise of improved download speed is consistently kept and if it warrants complexity and misbehavior with dups. If not consistent, then remove multi-threaded downloading.

AndreyNikiforov:
>> @menkej I am skeptical about multithreaded downloading from a single source [...] I would be curious to see some test results (single threaded vs multi threaded downloading)

> I run some test downloads. I am on a 100 MBit connection. No clear lab conditions - Netflix and Skype session running. Downloading 1000 recent files.

> default setting (multithreaded)
> 87,59s user 30,36s system 11% cpu 16:33,71 total
> 77,79s user 27,50s system 12% cpu 13:58,24 total
> 77,67s user 27,30s system 13% cpu 12:34,12 total

> --threads-num 1
> 79,00s user 27,64s system 14% cpu 12:34,61 total
> 79,17s user 27,51s system 11% cpu 15:04,41 total
> 78,30s user 27,80s system 14% cpu 12:36,18 total

> Just three runs. Just 1000 pics. Seeing not much of a difference really.

> Multithreading might make sense for parallel image/video conversions and stuff. For download... not so sure...

> Maybe it makes a difference on a clear line with a big pipe...

> _Originally posted by @menkej in https://github.com/ndbroadbent/icloud_photos_downloader/pull/171#issuecomment-710601811_

PR #171, opened 2020-10-10
improve test coverage

AndreyNikiforov:

PR #191, opened 2020-10-27
remove multi-threaded downloading

AndreyNikiforov:
feat: removed multi-threaded downloading and added deprecation notice to --threads-num parameter #180, #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ pip install icloudpd
                                         prints log messages on separate lines
                                         (Progress bar is disabled by default if
                                         there is no tty attached)
-        --threads-num INTEGER RANGE     Number of cpu threads (default: 1)
+        --threads-num INTEGER RANGE     Number of cpu threads -- depricated. To be removed in future version
         --version                       Show the version and exit.
         -h, --help                      Show this message and exit.
 
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ pip install icloudpd
                                         prints log messages on separate lines
                                         (Progress bar is disabled by default if
                                         there is no tty attached)
-        --threads-num INTEGER RANGE     Number of cpu threads -- depricated. To be removed in future version
+        --threads-num INTEGER RANGE     Number of cpu threads -- deprecated. To be removed in future version
         --version                       Show the version and exit.
         -h, --help                      Show this message and exit.
 
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -188,7 +188,7 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
               )
 @click.option(
     "--threads-num",
-    help="Number of cpu threads (default: 1)",
+    help="Number of cpu threads -- depricated. To be removed in future version",
     type=click.IntRange(1),
     default=1,
 )
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -188,7 +188,7 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
               )
 @click.option(
     "--threads-num",
-    help="Number of cpu threads -- depricated. To be removed in future version",
+    help="Number of cpu threads -- deprecated. To be removed in future version",
     type=click.IntRange(1),
     default=1,
 )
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -118,7 +118,7 @@ pip install icloudpd
                                         prints log messages on separate lines
                                         (Progress bar is disabled by default if
                                         there is no tty attached)
-        --threads-num INTEGER RANGE     Number of cpu threads (default: 1)
+        --threads-num INTEGER RANGE     Number of cpu threads -- depricated. To be removed in future version
```
ndbroadbent, line 121: Minor typo here - should be "deprecated"

icloudpd/base.py
```diff
@@ -194,7 +188,7 @@
               )
 @click.option(
     "--threads-num",
-    help="Number of cpu threads (default: 1)",
+    help="Number of cpu threads -- depricated. To be removed in future version",
```
ndbroadbent, line 191: also here "deprecated"

ndbroadbent:
Sure, I just saw this follow-up comment about the download times: https://github.com/icloud-photos-downloader/icloud_photos_downloader/pull/171#issuecomment-715300917

So that sounds fine to me!

tests/test_download_photos.py
```diff
@@ -322,7 +322,6 @@ def test_until_found(self):
         files_to_skip.append(("2018/07/30/IMG_7400.JPG", "photo", 2308885))
         files_to_skip.append(("2018/07/30/IMG_7400-medium.MOV", "photo", 1238639))
         files_to_skip.append(("2018/07/30/IMG_7399.JPG", "photo", 2251047))
-        files_to_download.append(("2018/07/30/IMG_7399-medium.MOV", "photo"))   # TODO this should not be downloaded, correct?
```
menkej, line 325: Why did you remove this? Is that the file that causes the test issues for you?

menkej (commented):
Fine with me - apart from the typo, but I'm currently running some tests on multithreading, so let's wait some days before we really remove it.

tests/test_download_photos.py
AndreyNikiforov: That file should never be downloaded, because all processing should stop after three skips. i added check to the test.

README.md
AndreyNikiforov: fixed

AndreyNikiforov:
> Fine with me - apart from the typo, but I'm currently running some tests on multithreading, so let's wait some days before we really remove it.

Current (1.6.2) multi-threaded downloading implementation has a number of shortcomings that cause a majority of issue reported for the project since I started paying more attention to it:
- deduplication does not work with threads > 1, we changed default to 1 to reduce chances of ppl hitting this issue
- until_found is inconsistent (btw it forces threads == 1), causing failing tests (==false alarms); this is because current pub-sub implementation checks counter at publisher and moving this check to subscriber requires additional back pressure implementation

We did confirm in a number of tests that multi threaded downloading does not improve speed on fast and slow connections. Anybody should be able to repeat tests and validate for themselves using 1.6.2. Here are things to keep in mind while testing multi-threaded vs single-threaded:
- iCloud account should not have any duplicates
- test should not use --unit-found parameter

If we are to fix/re-implement parallel downloading, we'll have to keep the following in mind:
- deduplication should take both in-flight tasks and files on disk into account
- until-found may be really tricky to implement to avoid any unnecessary processing/downloading
- it would be most likely more resource efficient to use event loop-based IO concurrency like asyncio

With all that in mind, I feel it would be better to release this feature as 1.7.0, continue any multi-threaded downloading testing using 1.6.2, and revisit parallel downloading if tests show any promises and we have resources to implement parallel downloading correctly. @menkej will that work for you?

jmonster:
@AndreyNikiforov was this feature re-added? The cli arg `--help` still advertises this feature (as does the README). I'm not asking you to put it back, but if it is _is_ supposed to work, it doesn't seems broken

https://github.com/icloud-photos-downloader/icloud_photos_downloader/search?q=threads_num

AndreyNikiforov:
> was this feature re-added? 

What makes you think so?

> The cli arg `--help` still advertises this feature (as does the README). I'm not asking you to put it back, but if it is _is_ supposed to work, it doesn't seems broken

What do you mean by `advertising`? Help in 1.7.2 says that feature is deprecated and will be removed. It is not broken, it is just noop (==does nothing).

Issue #192, opened 2020-10-27
Failed to run docker on Windows


AndreyNikiforov:
```
docker run -it --rm --name icloud_andrey -e TZ=America/Los_Angeles -v m:/photos/andrey:/data -v m:/iCloudCookies:/cookies icloudpd/icloudpd:1.6.2-alpine icloudpd --directory /data --cookie-directory /cookies --username REDACTED --folder-structure {:%Y/%Y-%m-%d} --auto-delete --threads-num 1 --no-progress-bar --until-found 50
Unable to find image 'icloudpd/icloudpd:1.6.2-alpine' locally
1.6.2-alpine: Pulling from icloudpd/icloudpd
Digest: sha256:20ec1dcd7282fdde3160e04bc2313b918e37c568632e3ba772a345bcafa085b5
Status: Downloaded newer image for icloudpd/icloudpd:1.6.2-alpine
2020-10-27 19:39:36 DEBUG    Authenticating...
iCloud Password: 
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/icloudpd/authentication.py", line 29, in authenticate
    icloud = pyicloud_ipd.PyiCloudService(
  File "/usr/lib/python3.8/site-packages/pyicloud_ipd/base.py", line 141, in __init__
    password = get_password_from_keyring(apple_id)
  File "/usr/lib/python3.8/site-packages/pyicloud_ipd/utils.py", line 40, in get_password_from_keyring
    raise NoStoredPasswordAvailable(
pyicloud_ipd.exceptions.NoStoredPasswordAvailable: No pyicloud password for REDACTED could be found in the system keychain.  Use the `--store-in-keyring` command-line option for storing a password for this username.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/icloudpd", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3.8/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
[...]

PR #193, opened 2020-10-27
remove USER from Dockerfile to avoid permission issues

AndreyNikiforov:
fixed alpine image permissions on Windows #192

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -10,11 +10,11 @@ RUN set -xe && \
     rm -r /root/.cache
 
 ARG ICLOUDPD_VERSION
-ADD icloudpd-latest-py2.py3-none-any.whl /tmp
+COPY dist/* /tmp
 RUN set -xe \
-  && pip install /tmp/icloudpd-latest-py2.py3-none-any.whl \
+  && pip install /tmp/icloudpd-${ICLOUDPD_VERSION}-py2.py3-none-any.whl \
   && icloudpd --version \
   && icloud -h | head -n1
 
-RUN adduser -D -h /home/user -u 1000 user
-USER user
+# RUN adduser -D -h /home/user -u 1000 user
+# USER user
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 build/
-dist/
 htmlcov/
 icloudpd.egg-info
 cron_script.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- fix: docker works on Windows #192
+
 ## 1.6.2 (2020-10-23)
 
 - Began recording updates in `CHANGELOG.md`
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,1 +1,2 @@
 twine
+wheel
--- a/scripts/build_docker
+++ b/scripts/build_docker
@@ -6,7 +6,6 @@ ROOT_DIR="$(realpath $(dirname "$0")/..)"
 ICLOUDPD_VERSION="$(cat "$ROOT_DIR/setup.py" | grep version= | cut -d'"' -f 2)"
 echo "Current icloudpd version: ${ICLOUDPD_VERSION}"
 
-cp ./dist/icloudpd* ./icloudpd-latest-py2.py3-none-any.whl
 docker build \
   -t "icloudpd/icloudpd:${ICLOUDPD_VERSION}" \
   -f "Dockerfile.release" \


Dockerfile.release
```diff
@@ -10,11 +10,11 @@ RUN set -xe && \
     rm -r /root/.cache
 
 ARG ICLOUDPD_VERSION
-ADD icloudpd-latest-py2.py3-none-any.whl /tmp
```
menkej, line 13: I'm not really sure what this .whl is for, so I can't really say something to that change. Why are the user additions commented out? AFAIR we had an issue about user rights (cron, suid, something) in relation with docker... Is docker on other platforms still fine? Can we test this in any unit test?

menkej (commented):
Not the docker expert. Some of the changes look like they interfere with other platforms...

Dockerfile.release
boredazfcuk: I noticed this the other night too. I was going to try building, but couldn't find this file and presumed the Dockerfile build would not be able to find it.

Assumed this must exist on the machine that is building the Docker images.

AndreyNikiforov: We have two dockerfile`s in the repo:
- `Dockerfile` that uses python base image and uses current source tree to build image (including testing, pulling all dependencies etc); it produces *fat* image (1.7G from latest master).
- `Dockerfile.release` uses alpine base and relies on the rest of release scripts (test, lint, python package into .whl) for all the work except packaging docker image; it produces 70k image and this is the one that is published to dockerhub.

This PR is about fixes to `Dockerfile.release`.

To produce .whl locally, you need to run `scripts\build`

AndreyNikiforov: > I'm not really sure what this .whl is for, so I can't really say something to that change. Why are the user additions commented out? AFAIR we had an issue about user rights (cron, suid, something) in relation with docker... Is docker on other platforms still fine? Can we test this in any unit test?

Tested on Mac before and after the change -- both versions worked.
Windows broken before and worked after the change.
Linux -- did not test.

Testing docker automatically would be awesome:
1. need to figure out how to simulate icloud traffic, e.g. from existing vcr recordings.
1. need to figure out how GitHub actions work with docker
1. need to figure out how GitHub support different platforms and ideally architectures, so we can test Win/Mac/Lin and x64/arm

That is important, but separate enhancement to our dev lifecycle.

AndreyNikiforov: Tested on Linux (Ubuntu 20.04) before and after the change -- both versions worked.

AndreyNikiforov:
@ndbroadbent do you have feedback on this change?

Dockerfile.release
```diff
@@ -10,11 +10,11 @@ RUN set -xe && \
...
+COPY dist/* /tmp
```
ndbroadbent, line 13: The only issue with this change is that `dist/*` can contain multiple older versions, so this COPY command will include all of the older versions inside the `/tmp` folder in the Docker image. That's the reason why I was copying the latest specific version into the root folder, just to be careful about space and keep the Docker image as small as possible. But this can also be fixed by deleting all files in the `dist/*` folder before building, so I will update the build script to do that

ndbroadbent:
Sure I'm just really confused about why this change would make any difference on Windows. The actual Docker container should be identical, so I'm not sure why changing the USER line would make a difference depending on the host OS. 

I read through #192, and was wondering about this comment: https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/192#issuecomment-717532900

Does it work if we include the `keyrings.alt==1.0 keyring==8.0` dependencies?

boredazfcuk:
> Sure I'm just really confused about why this change would make any difference on Windows. The actual Docker container should be identical, so I'm not sure why changing the USER line would make a difference depending on the host OS.

I think it's down to the cookie location being mounted inside the container. Currently Andrey is mounting the container as follows:
```
docker run -it --rm --name icloud_andrey -e TZ=America/Los_Angeles -v m:/photos/andrey:/data -v m:/iCloudCookies:/cookies icloudpd/icloudpd:1.6.2-alpine icloudpd --directory /data --cookie-directory /cookies --username REDACTED --folder-structure {:%Y/%Y-%m-%d} --auto-delete --threads-num 1 --no-progress-bar --until-found 50
```
The cookie folder inside the container is being mounted to m:/iCloudCookies on the host machine. The filesystem permissions on the host machine will apply. 

Currently the container is set to run as user `user` with ID 1000. Presumably, this user doesn't have access to the m:/iCloudCookies location on a Windows host machine. When the `USER user` line is removed, the issue goes away as the container is now running and root. Presumably the root account does have permissions to the m:/iCloudCookie location on the host machine.

Chances are it will work on a Linux machine, because the primary user is 1000 and the docker volume will be mounted to a location that allows access for that ID. I have no experience with Mac, but presume it recognises UIDs in the same way as Linux.

So it's either that, or that it's because it's trying to use keyring based authentication when it doesn't have the required packages installed.

AndreyNikiforov:
> Sure I'm just really confused about why this change would make any difference on Windows. The actual Docker container should be identical, so I'm not sure why changing the USER line would make a difference depending on the host OS.
> 
> I read through #192, and was wondering about this comment: [#192 (comment)](https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/192#issuecomment-717532900)
> 
> Does it work if we include the `keyrings.alt==1.0 keyring==8.0` dependencies?

The error I got was repeated to permissions, so checking the code that adjusts user id (`USER` docker command) was natural first step for me. Since I did see neither use of the adjusted user id nor documentation (including commit history), I decided to try removing it. And it worked. I don't know how extra libraries could solve permission issues, but I did not try that approach assuming that removal of the `USER` cmd is the solution.

ndbroadbent:
Sure this should be fine! It's pretty normal to run things as root in Docker anyway, and it's run locally on a user's machine. So this is fine.

Dockerfile.release
AndreyNikiforov: `/dist` is ignored in git and removed in `scripts\clean`, so this concern has been addressed before this change

PR #194, opened 2020-10-28
Fix log-level option

ltm:
The log level is set on the logger instance in `icloudpd.base.main` based on the `--log-level` option, but subsequent calls to `setup_logger()` resets the level to `DEBUG`. As a result the `--log-level` option is effectively ignored.

Since the `--log-level` option defaults to `DEBUG` it's safe to remove `logger.setLevel()` from `setup_logger()` which allows the option to work as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix: --log-level option [#194](https://github.com/icloud-photos-downloader/icloud_photos_downloader/pull/194)
 - feature: Folder structure can be set to 'none' instead of a date pattern,
 so all photos will be placed directly into the download directory.
 - fix: Empty directory structure being created #185
--- a/icloudpd/logger.py
+++ b/icloudpd/logger.py
@@ -2,7 +2,7 @@
 
 import sys
 import logging
-from logging import DEBUG, INFO
+from logging import INFO
 
 
 class IPDLogger(logging.Logger):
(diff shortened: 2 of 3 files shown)


menkej:
Hi @ltm !
Thanks for your pull request! The change looks rather simple. I'll test it later.

Can you also add a test case that checks the logging handling?

ltm:
@menkej I've updated the existing `test_log_levels` test to check for the appropriate levels in the output. For some reason an unrelated test (i.e. `test_until_found`) has started failing, but I see it's failing in other PRs as well so I think it's unrelated to my changes.

menkej (changes requested):
Thanks updating the test cases! I'l request one additional check: Both tests assume the logger running on INFO. There should also be a test for DEBUG level as well, what do you think?

menkej:
test_until_found is an issue of its own. It behaves inconsistent (I think we have a PR for that already). I have no Idea why it sometimes runs fine and sometimes does not...

ltm:
@menkej Makes sense. I've updated (and refactored) the test to check all three log levels.

menkej (commented):
Nice work @ltm! Way better than before. Last little cream on the cake is to add this fix to the CHANGELOG.md in the unreleased section as "fix:" and then we're ready to merge! Sorry I forgot to mention this before. We just introduced the CHANGELOG recently and I'm still not used to it... ;-)

ltm:
@menkej No problem. I've added a changelog entry.

menkej (approved):

PR #169, opened 2020-10-09
Allow folder-structure to be none

sinaxin:
When set to none, allows to download all photos to base directory.

PR #195, opened 2020-10-28
Folder none

menkej:
Recreation of pull request #169 because uncertainty about correct merging. Initial idea by @sinaxin.


menkej:
Ok, createrelease-Action also ended up here... new actions in GitHub are not so easily started in branches... :-( Should be of no harm anyways and can be disabled. This action creates a new Release on Github if a v*-Tag, like v1.6.3 is created... Just an experiment in context of easier/more automated release process.

.github/workflows/createrelease.yml
```diff
@@ -0,0 +1,39 @@
+on:
```
AndreyNikiforov, line 1: Is this an artifact of rebase?

icloudpd/base.py
```diff
@@ -390,7 +391,10 @@ def download_photo(counter, photo):
             created_date = photo.created
 
         try:
-            date_path = folder_structure.format(created_date)
+            if folder_structure == "none":
```
AndreyNikiforov, line 394: Is there a way to make this search case insensitive?

AndreyNikiforov (changes requested):
Added couple comments/suggestions. Would be nice to address them.

.github/workflows/createrelease.yml
menkej: My master was just this commit ahead of upstream/master. If you create a new Action they always end up in master. I will disable it in upstream straight after merge, ok?

AndreyNikiforov: I see. What are the reasons to have master in the fork ahead of the upstream master? Is that because GitHub release actions are not working from branches?

icloudpd/base.py
menkej: good point.

.github/workflows/createrelease.yml
menkej: At least that is what I experienced. I  created a branch for the actions, switched to that branch in the web interface, created the action, it ended up in master... This was not what I expected...

menkej: forked master should not be ahead of upstream, I totally agree on that. I'm still learning all those GitHub features, so I don't want to rule out I made a mistake somewhere with these Actions... If I disable it right after merge it will not interfere with anything, though.

menkej: Or should we just delete the file here?

AndreyNikiforov: If file is not in upstream master, then deleting it in PR will:
1. make PR specific to the change (no other unrelated changes)
1. safe to merge (if file is in upstream master, merge may delete it AFAIK)

PR #201, opened 2020-11-02
fixes #200, --only-print-filenames (again)

### LINKED ISSUE #200: --only-print-filenames option displays filenames of files that have already been downloaded
Hi,

I've noticed that if you use the `--only-print-filenames` option to display the list of files that will be downloaded, it shows files which are already on disk:

```
/ $ /usr/bin/icloudpd --directory /home/user/iCloud --cookie-directory /config --username notmy@email.com --folder-structure {:%Y/%m} --only-print-filenames
/home/user/iCloud/2020/09/IMG_1693_HEVC.MOV
/home/user/iCloud/2020/09/IMG_1691_HEVC.MOV
/home/user/iCloud/2020/09/IMG_1685_HEVC.MOV
/home/user/iCloud/2020/06/IMG_1646_HEVC.MOV
/home/user/iCloud/2020/05/IMG_1606_HEVC.MOV
/home/user/iCloud/2020/05/IMG_1605_HEVC.MOV
/home/user/iCloud/2020/05/IMG_1577_HEVC.MOV
/home/user/iCloud/2019/03/IMG_0331_HEVC.MOV
/home/user/iCloud/2019/03/IMG_0245_HEVC.MOV
/home/user/iCloud/2019/03/IMG_0244_HEVC.MOV
/ $ ls /home/user/iCloud/2020/09/IMG_16*
IMG_1685.HEIC      IMG_1687.MOV       IMG_1691.HEIC      IMG_1692.PNG       IMG_1693_HEVC.MOV  IMG_1696.PNG       IMG_1699.PNG
IMG_1685.JPG       IMG_1688.MOV       IMG_1691.JPG       IMG_1693.HEIC      IMG_1694.PNG       IMG_1697.PNG
IMG_1685_HEVC.MOV  IMG_1689.MOV       IMG_1691_HEVC.MOV  IMG_1693.JPG       IMG_1695.PNG       IMG_1698.PNG
```

This option should return an empty list if all files are already downloaded. However, it does not.

I've noticed that all of the above files are 'Live Photos', if that helps?

Thanks.

menkej:
Just a one-liner, so that Live Photos are not tried to download again... (and printed...)

### DIFF
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -516,7 +516,7 @@ def main(
 
                 lp_file_exists = os.path.isfile(lp_download_path)
 
-                if only_print_filenames:
+                if only_print_filenames and not lp_file_exists:
                     print(lp_download_path)
                 else:
                     if lp_file_exists:
--- a/tests/test_download_live_photos.py
+++ b/tests/test_download_live_photos.py
@@ -1,0 +1,195 @@
+from unittest import TestCase
+from vcr import VCR
+import os
+import sys
+import shutil
+import pytest
+import mock
+import datetime
+from mock import call, ANY
+from click.testing import CliRunner
+import piexif
+from piexif._exceptions import InvalidImageDataError
+from pyicloud_ipd.services.photos import PhotoAsset, PhotoAlbum, PhotosService
+from pyicloud_ipd.base import PyiCloudService
+from pyicloud_ipd.exceptions import PyiCloudAPIResponseError
+from requests.exceptions import ConnectionError
+from icloudpd.base import main
+from tests.helpers.print_result_exception import print_result_exception
+
+vcr = VCR(decode_compressed_response=True)
+
+class DownloadLivePhotoTestCase(TestCase):
+    @pytest.fixture(autouse=True)
+    def inject_fixtures(self, caplog):
+        self._caplog = caplog
+
+    def test_skip_existing_downloads_for_live_photos(self):
+        base_dir = os.path.normpath("tests/fixtures/Photos")
+        if os.path.exists(base_dir):
+            shutil.rmtree(base_dir)
+        os.makedirs(base_dir)
+
+        with vcr.use_cassette("tests/vcr_cassettes/download_live_photos.yml"):
+            # Pass fixed client ID via environment variable
+            os.environ["CLIENT_ID"] = "DE309E26-942E-11E8-92F5-14109FE0B321"
+            runner = CliRunner()
+            result = runner.invoke(
+                main,
+                [
+                    "--username",
+                    "jdoe@gmail.com",
+                    "--password",
+                    "password1",
+                    "--recent",
+                    "3",
+                    "--no-progress-bar",
+                    "--threads-num",
+                    1,
+                    "-d",
+                    base_dir,
+                ],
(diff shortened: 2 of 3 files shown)


icloudpd/base.py
```diff
@@ -516,7 +516,7 @@ def download_photo(counter, photo):
 
                 lp_file_exists = os.path.isfile(lp_download_path)
 
-                if only_print_filenames:
+                if only_print_filenames and not lp_file_exists:
```
AndreyNikiforov, line 519: We need a test for this case.

Quickly scanning through the logic tree, I suspect that executing icloudpd with `--only-print-files` over duplicate live photo assets with different sizes where one already had been downloaded will cause unexpected side effects.

AndreyNikiforov (changes requested):

icloudpd/base.py
menkej: I agree, but I did not come up with a nice test scenario... Ideas? I'm still not well up with these cassettes and stuff.

Actually I don't really like the structure in that section at all. Why that extra handling for Live fotos. Should work more generic and be refactored somewhen...

menkej: Tests added. This was far more work than the fix itself... ;-) But a good invest in (future) quality.

tests/test_download_live_photos.py
```diff
@@ -0,0 +1,195 @@
...
+                os.path.join(base_dir, os.path.normpath("2020/11/04/IMG_0512.HEIC")),
+                filenames[1]
+            )
+            self.assertEqual(
+                os.path.join(base_dir, os.path.normpath("2020/11/04/IMG_0512_HEVC.MOV")),
+                filenames[2]
+            )
+
+            # Double check that a mocked file does not get listed again. Its already there!
+            assert "2020/11/04/IMG_0514_HEVC.MOV" not in filenames
```
AndreyNikiforov, line 193: Dedup logic will try to download live photo with `-<size>` suffix, correct? If yes, then this assert will not catch that branch of the logic.

menkej: I hope I get your remark right. Line 193 is not needed to test this bugfix, as if the live-photo (.mov) would get listet for download again (which was the initial bug but does not happen after the fix, of course) the order of the first entries in the file-list changes and one of the 3 checks before already complains (maybe I should have been more descriptive in the comment in line 192). If you use --live-photo-size the bugfix logic chatches that as well, als the check for existence is right before the bugfix.

So I see no need for changing. Or did I miss something?

AndreyNikiforov: I am talking about `2020/11/04/IMG_0514_HEVC-3951774.MOV` file that will be still downloaded if `2020/11/04/IMG_0514_HEVC.MOV` exists and `--only-print-filenames` selected. Checking the full output of the processing logic will reveal that fact.

menkej: Exactly. I mocked the .mov intentionally and expect the photo to be downloaded/listed.

AndreyNikiforov: You are right, my case is edge

AndreyNikiforov (approved):

PR #212, opened 2020-11-05
improve layer caching in docker

AndreyNikiforov:
This is minor improvement to reuse cached docker layers. No change log because it is not consumer facing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,17 +5,19 @@ FROM python:3.9 as base
 WORKDIR /app
 COPY requirements.txt .
 RUN pip install -r requirements.txt
-COPY . .
 
 FROM base as test
 
 RUN mkdir Photos
+COPY requirements-test.txt .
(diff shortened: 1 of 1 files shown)


menkej (approved):
I'm not really into docker yet, so I have no opinion. What would `COPY . .` do? If you can enlighten me in 2 sentences (no effort) that would be nice. Otherwise I will somewhen dive into this myself. ;-)

AndreyNikiforov:
> I'm not really into docker yet, so I have no opinion. What would `COPY . .` do? If you can enlighten me in 2 sentences (no effort) that would be nice. Otherwise I will somewhen dive into this myself. ;-)

`COPY . .` copies all files (except `.dockerignore`d) from working folder into container. Moving this command after installing requirements gives a chance for docker to reuse cached file system layer with requirements from previous builds when some of the files in the current folder change.

PR #213, opened 2020-11-05
pin dev and test dependencies

AndreyNikiforov:
Pin dependency version for dev and test requirements. Dependabot will handle explicit upgrades. No change log because it is not consumer facing.


menkej (commented):
I'm not really sure why we should pin these versions...

At least we should open up the patch-level, don't you think? That would be `pkg>=1.1,<=1.2` or something I guess.

AndreyNikiforov:
> I'm not really sure why we should pin these versions...
> 
> At least we should open up the patch-level, don't you think? That would be `pkg>=1.1,<=1.2` or something I guess.

Pinning significantly improves determinism in our processes (easy to reproduce issues, lower/no chance of sporadic breaks). The downside of the pinning is an increase in maintenance, but with debendabot it is significantly reduced, so we can have determinism for the low cost. I agree that patch updates have lower risks, but I hope that with dependadot these updates will be painless.

AndreyNikiforov:
Expecting to fix #216 by downgrading dependencies. Also reusing scripts to install all dependencies and test, so we can catch bad upgrades

menkej (approved):
Fine with me for now, even though I fear that in the end we pin everything and outdate over time... And still I think it would be better just to pin on minor not on explicit path versions.

AndreyNikiforov:
> Fine with me for now, even though I fear that in the end we pin everything and outdate over time... And still I think it would be better just to pin on minor not on explicit path versions.

We'll see over next few months how dependabot handles updates and then decide. With testing as github actions should be able to catch any bad upgrades.

PR #215, opened 2020-11-08
skip problematic tests on mac

AndreyNikiforov:
date handling tests do not work on win and mac. skipping them until we have time to dig deeper.

all tests are passing (or skipping) on all platforms now


menkej (approved):
We can do this but should file an issue to fix the skipped tests right away, so that this is not forgotten forever. The tests should be portable as well or have platform specific pendants...

Issue #181, opened 2020-10-21
Improving Release process


AndreyNikiforov:
Current process is a set of manual steps -- automating these will reduce chances of errors and reduce inconsistencies: today we have 1.6.0 release version in codebase, 1.6.1 in PyPy, 1.3.4 in GitHub, `latest` in dockerhub with 1.6.0

Current release process is performed by @ndbroadbent (Owner; the only one with access to pypy and dockerhub), creating potential delays for releases. Distributing this role among Maintainers will reduce dependencies.

I am proposing to use github actions for creating .whl assets, publishing PyPy packages, and pushing Docker images to dockerhub on each gitHub Release publishing

Benefits:
- All credentials to PyPy and Dockerhub remain in the hands of Owner (but have to be added to the repo secrets)
- Github Releases can be created by Maintainers, producing releases at any schedule
- Process is automated, consistent, and repeatable

Additionally, we can use github action for testing pull requests/commits instead of travis. Travis account will no longer be needed

Not changing:
- coverall

Challenges:
- implementing/testing gitlab actions may be challenging without publishing new versions

menkej:
I'm really interested in this. I'm having similar issues at my "day job", where we're just starting to get near to what you describe @AndreyNikiforov. It makes totally sense for me. I have no glue of the publishing processes of these platforms but I'd be interested to dive into it.

menkej:
> we can use github action for testing pull requests/commits

I created a GitHub action for running the tests. Seems to work. Feel free to mess around with it. ;-)

AndreyNikiforov:
> > we can use github action for testing pull requests/commits
> 
> I created a GitHub action for running the tests. Seems to work. Feel free to mess around with it. ;-)

@menkej That is very nice work! Follow up questions and suggestions:
- Actions seems to be failing for `master` now. Can you see if they can be fixed?
- It may be easier to iterate on your own fork and then submit PR
- Do you think we can use testing on `push`es to every branch, not only `master`? That way workflow will work in forks and help maintaining quality before submitting PRs.

AndreyNikiforov:
@menkej another ideas provoked by your testing action: use matrix strategy to test on different python versions (https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix)

menkej:
> > > we can use github action for testing pull requests/commits
> > 
> > 
[...]

PR #221, opened 2020-11-08
fix script to build release docker

AndreyNikiforov:
should help with #181


menkej (approved):
Fine with me. I had made almost the exact changes while creating the release action. Let's see how git handles this duplication. ;-)

PR #223, opened 2020-11-09
Automation for release process

### LINKED ISSUE #181: Improving Release process

menkej:
- new release is automatically created when a new v*-tag is created
- version v* must match setup.py
- Releasenotes are taken from CHANGELOG.md
- Python package is build and pushed to pypy
- Docker image is build (based on Dockerfile.release) and pushed to dockerhub

This closes #181 (apart from the also there discussed improvement to Dockerfile.release suggested by @boredazfcuk)

### DIFF
--- a/.github/workflows/createrelease.yml
+++ b/.github/workflows/createrelease.yml
@@ -1,0 +1,99 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Create Release
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Retrieve version and Changelog
+        run: |
+          echo ICLOUDPD_VERSION=$(cat setup.py | grep version= | cut -d'"' -f 2) >> $GITHUB_ENV
+          echo "::debug::${{env.ICLOUDPD_VERSION}}"
+          echo 'ICLOUDPD_CHANGELOG<<EOF' >> $GITHUB_ENV
+          scripts/extract_releasenotes CHANGELOG.md >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+          echo "::debug::${{env.ICLOUDPD_CHANGELOG}}"
+      - name: Install dependencies
+        # check if the tag matches the version in setup.py
+        if: endsWith(github.ref, env.ICLOUDPD_VERSION)
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install setuptools wheel twine
+      - name: Build and publish
+        if: endsWith(github.ref, env.ICLOUDPD_VERSION)
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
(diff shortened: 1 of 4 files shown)


menkej:
This workflow still creates a universal wheel. I did not change this as `scripts/build` does the same at the moment. As we decided to drop support for Python 2 we should not use universal in the future as well. What do you think?

.github/workflows/createrelease.yml
```diff
@@ -0,0 +1,99 @@
...
+          echo 'ICLOUDPD_CHANGELOG<<EOF' >> $GITHUB_ENV
+          scripts/extract_releasenotes CHANGELOG.md >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+          echo "::debug::${{env.ICLOUDPD_CHANGELOG}}"
+      - name: Install dependencies
+        # check if the tag matches the version in setup.py
+        if: endsWith(github.ref, env.ICLOUDPD_VERSION)
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install setuptools wheel twine
```
AndreyNikiforov, line 29: We need to use `scripts/install_deps` because pulling latest breaks dependency for libs. That script will cover pip upgrade too.

Reusing existing scripts in other steps is not as critical and can be refactored later.

AndreyNikiforov (changes requested):
Awesome work!

How can we handle failures and specifically retries after failure? For example if push to dockerhub fails, we'll need to kick the whole action again. Uploading to Pypi may be idempotent, and creating github release does not sound like it. This concern is not a blocker for PR as this topic will require more research and iterating.

.github/workflows/createrelease.yml
menkej: > How can we handle failures and specifically retries after failure? For example if push to dockerhub fails, we'll need to kick the whole action again. Uploading to Pypi may be idempotent, and creating github release does not sound like it. This concern is not a blocker for PR as this topic will require more research and iterating.

Up till now that is a manual process. The steps build on each other. Worst: upload to docker fails: If you want to stick with the same release-version you would have to delete the release on pypi and on GitHub (including the tag - just 3 clicks) and then recreate the tag.

In total that is a matter of < 5 mins. If pypi upload fails its just about deleting the release on GitHub.

It should be possible to create an Action that deletes a possible upload in pypi and Dockerhub if a Release is deleted. But for the moment I don't think that's worth the effort...

AndreyNikiforov (approved):

PR #224, opened 2020-11-10
Fixup dependencies and prepare new release

menkej:


.github/workflows/createrelease.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         # check if the tag matches the version in setup.py
         if: endsWith(github.ref, env.ICLOUDPD_VERSION)
         run: |
+          python -m pip install --upgrade pip
```
AndreyNikiforov, line 28: This is part of `scripts/install_deps` with explicit version to force failed dependencies to fail. We need to use one from the script

menkej: Actually I switched to the script and that led to failing! This is why I reverted. You can see the version history or oder runs of earlier today. After some trial&error I guess it was a python2 python3 issue. That's why I changed to explicit python3-commands. Currently the action uses the following:

```
python3 -m pip install --upgrade pip
python3 -m pip install setuptools
# this section should be updated. p.e. wheel is installed too late.
# this leads to unnecessary warnings
scripts/install_deps
```

Refer to `.github/workflows/createrelease.yml` on `master`. As I stated in the comment we need to work on that section.

PR #226, opened 2020-11-10
add validation build and explicit deps

AndreyNikiforov:
use explicit dependencies for all libs and pip -- that will allow deterministic testing, including dependabot upgrade suggestions.

adding validation build as part of the quality checks, so we are sure that our release pipeline is executed on each push (except publishing). should help polishing release actions #224


Dockerfile
```diff
@@ -3,14 +3,16 @@
 FROM python:3.9 as base
 
 WORKDIR /app
+# explicit requirements because runtime does not need ALL dependencies
+COPY requirements-pip.txt .
 COPY requirements.txt .
-RUN pip install -r requirements.txt
+RUN pip install -r requirements-pip.txt
```
menkej, line 9: Is this a new file? Missing?

menkej (changes requested):
Seems like a new file is introduced that is missing in the pull

Dockerfile
AndreyNikiforov: hmm. shows in the list of files in github UI for me

requirements.txt
```diff
@@ -1,5 +1,4 @@
 pyicloud_ipd==0.10.1
-#docopt==0.6.2
```
menkej, line 2: I think that was a relic from the past anyways.

requirements-dev.txt
```diff
@@ -1,3 +1,3 @@
+setuptools==50.3.2
```
menkej, line 1: I still think we should only pin on minor releases... ;-)

.github/workflows/quality-checks.yml
```diff
@@ -35,3 +35,45 @@ jobs:
...
+      uses: docker/setup-qemu-action@v1
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Build
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./Dockerfile
```
menkej, line 77: We are building releases with Dockerfile.release. Testing now with Dockerfile. Should this run `--target=test`?

What should happen here?

AndreyNikiforov: Good point. Adding release docker validation

requirements-dev.txt
AndreyNikiforov: With dependabot the cost of upgrade is close to zero and pinning to patch version guarantees ~100% reproduction of the process.

menkej:
while you're on it: what do you think about shortening the name of the jobs? `quality_checks_ubuntu_python`-> we know its ubuntu and python, so `quality_checks` is sufficient imho. shortens the tabs on the left in "Actions". What do you think?

AndreyNikiforov:
> while you're on it: what do you think about shortening the name of the jobs? `quality_checks_ubuntu_python`-> we know its ubuntu and python, so `quality_checks` is sufficient imho. shortens the tabs on the left in "Actions". What do you think?

done

menkej (approved):

PR #228, opened 2020-11-11
run tests in parallel

AndreyNikiforov:
in github running tests in parallel ends 2x faster (3m instead of 7m). Forces developers keeping all tests without shared resources, e.g. common writable folder. No change log as this is not customer facing

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,6 @@
+.DS_Store
 build/
+dist/
 htmlcov/
 icloudpd.egg-info
 cron_script.sh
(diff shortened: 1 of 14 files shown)


menkej (approved):
Great idea! Definitely worth a try.

But a t least one comment about why the folder is created like this might help new devs... So another thing wo should loose a word on in a little "how to write a test" how-to.

Submitted just 1h ago. What time zone are you on? Do you sleep? ;-)

menkej:
Now I finally learned how to test a pull locally.... It definitely also speeds up the test-run on my good old 2015 MBP! But it also burns the cpu... ;-) Good thing! I'm just wondering about a parameter for scripts/test to manually specify the limit (and keep the computer usable...). Maybe I'll add that later.

PR #232, opened 2020-11-14
fix docker build on win

AndreyNikiforov:
Linux shell is sensitive to line ends in scripts. When scripts are executed inside dev docker build, they must match used image os (linux).

Git is smart to convert line ends for text files depending on the platform. Docker builds on macos or linux get correct line ends from the repo. When repo is on windows, git checks out scripts as crlf (==windows) and docker build copies them to the image blindly, causing shell execution error.

As a fix, we use dos2unix at the build step to ensure that scripts are always with unix line ends inside docker.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-*.sh text eol=lf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: dev docker build on windows correctly manages crlf for scripts
+
 ## 1.7.1 (2020-11-10)
 
 - fix: --only-print-filenames option displays filenames (live photos) of files that have already been downloaded #200
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,14 @@ RUN pip3 install --use-feature=2020-resolver -r requirements.txt
 
 FROM base as test
 
+RUN apt-get update && apt-get install -y dos2unix
 RUN mkdir Photos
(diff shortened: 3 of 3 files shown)


menkej:
So how I understood it: on Windows/Mac everything is fine. If you work on a Windows checkout the files are checked out with windows line-endings (^M) and so the docker build would just stupidly copy the files to the image (with win endings) and they don't work any more? So in reality the files are OK its just the checkout on windows that breaks it? In my checkout on Mac all files are "clean" unix ffm (according to vim).

AndreyNikiforov:
> So how I understood it: on Windows/Mac everything is fine. If you work on a Windows checkout the files are checked out with windows line-endings (^M) and so the docker build would just stupidly copy the files to the image (with win endings) and they don't work any more? So in reality the files are OK its just the checkout on windows that breaks it? In my checkout on Mac all files are "clean" unix ffm (according to vim).

you are close. on **Linux** (not Windows) and Mac everything is fine -- git checks out text files with line endings matching platform. On windows git does the right thing too -- it checks out line ends matching platform (==crlf). Docker build for dev docker (==Dockerfile, not Dockerfile.release) copies text files without any conversion, so on Linux and Mac scripts appear inside docker image with line endings matching expected os (==linux inside docker). On Windows it causes issue.

BTW No way to test this in quality checks as GitHub docker builds are Linux-only actions.

menkej:
> you are close. on **Linux** (not Windows) and Mac everything is fine

This was exactly what I meant to type... ;-)

Issue #243, opened 2020-11-28
OverflowError when downloading photo


ian-mcdowell:
I'm unable to successfully run the script. It logs a bunch of photo names from 2020 back until 2015, then crashes with an error. There must be a photo in my library with a created at date that is invalid in some way.

```
...
/home/user/iCloud/2015/06/11/2015_06_11_183510.jpg
/home/user/iCloud/2015/06/11/2015_06_11_183503.jpg
/home/user/iCloud/2015/06/11/2015_06_11_181757.jpg
/home/user/iCloud/2015/06/11/2015_06_11_181457.jpg
Traceback (most recent call last):
File "/usr/bin/icloudpd", line 11, in <module>
load_entry_point('icloudpd==1.7.0', 'console_scripts', 'icloudpd')()
File "/usr/lib/python3.8/site-packages/click/core.py", line 722, in __call__
return self.main(*args, **kwargs)
File "/usr/lib/python3.8/site-packages/click/core.py", line 697, in main
rv = self.invoke(ctx)
File "/usr/lib/python3.8/site-packages/click/core.py", line 895, in invoke
return ctx.invoke(self.callback, **ctx.params)
File "/usr/lib/python3.8/site-packages/click/core.py", line 535, in invoke
return callback(*args, **kwargs)
File "/usr/lib/python3.8/site-packages/icloudpd-1.7.0-py3.8.egg/icloudpd/base.py", line 562, in main
File "/usr/lib/python3.8/site-packages/icloudpd-1.7.0-py3.8.egg/icloudpd/base.py", line 380, in download_photo
File "/usr/lib/python3.8/site-packages/pytz/tzinfo.py", line 201, in fromutc
return (dt + inf[0]).replace(tzinfo=self._tzinfos[inf])
OverflowError: date value out of range
```

menkej:
What are the exact options you're using to run the script?

IMcD23:
I'm running this in the latest build of the docker container from [docker-icloudpd](https://github.com/boredazfcuk/docker-icloudpd). 

Here are the logs from the top of the container's output:
```
2020-11-28 05:33:53 INFO ***** boredazfcuk/icloudpd container for icloud_photo_downloader started *****
2020-11-28 05:33:53 INFO ***** /usr/local/bin/sync-icloud.sh script version: 2020/11/21_15:10 *****

2020-11-28 05:33:53 INFO Alpine Linux v3.12
[...]

PR #245, opened 2020-11-29
Handle OverflowError when downloading photos with invalid dates

ian-mcdowell:
Addresses issue #243

### DIFF
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -378,7 +378,7 @@ def main(
             return
         try:
             created_date = photo.created.astimezone(get_localzone())
-        except (ValueError, OSError):
+        except (ValueError, OSError, OverflowError):
             logger.set_tqdm_description(
                 "Could not convert photo created date to local timezone (%s)" %
                 photo.created, logging.ERROR)


menkej:
Could you possibly add a test case for this situation? I don't have such an image in my iCloud so I can't reproduce it...

IMcD23:
I'm not familiar with your test code, and am not able to make more substantial changes here. 😞

menkej:
> I'm not familiar with your test code, and am not able to make more substantial changes here. 😞

Ok, I understand. For testing we use recorded (and anonymised communication with iCloud. If you'd be able to re-run a download (as small as possible using --recent xyz for example) we could use this as a basis for a unit test:

1. I could write a test case
1. you re-run it with your credentials, communication is recorded
1. you remove your personal data (email-address mostly) from the rest response and provide that recording
1. we could use that recording to retest this specific change and make sure this does not happen again if we do a refactoring or update third party libs.

What do you think? If you want to have a look: a similar test case would be in `tests/test_download_photos.py` in p.e. method `test_download_photos_and_set_exif_exceptions`. A sample of the recorded data we use can be found in `tests/vcr_cassettes/listing_photos.yml`

sensei-ishikawa:
this change did not work for me. i debbuged with pyCharm and time.mktime crashes when the year is 13. not sure why it is not assuming it is 2013.

i "hammered" the code and added a test for the year, if it falls outside of epoch, i just set it as the current date. totally not elegant, but functional.

def set_utime(download_path, created_date):
    """Set date & time of the file"""
    if created_date.year <= 1969:
        created_date = time.localtime()
    ctime = time.mktime(created_date.timetuple())
    os.utime(download_path, (ctime, ctime))


thanks folks for building this tool. it is great not to depend on the Photos app, which after the last 2 or 3 updates became a complete garbage. i have 100,000+ pictures and videos, Photos is unusable.

menkej:
> this change did not work for me. i debbuged with pyCharm and time.mktime crashes when the year is 13. not sure why it is not assuming it is 2013.

Thanks for you patch! Are you able to provide a test-case for this situation, including a cassette that can be used to reproduce this behaviour?

sensei-ishikawa:
actually, that crashed too with created_date.timetuple(), so i hard coded the date as epoch and i made the code even more awful. i am downloading 100,000+ items currently at 38%. 

how do i generate the cassete? i can reproduce the date object that causes the crash (tm_year = 13) and work on a better code to transform date into epoch. why aren't you using datetime? i don't recall ever using time.

def set_utime(download_path, created_date):
    """Set date & time of the file"""
    if created_date.year <= 1969:
        ctime = 1609396107
    else:
        ctime = time.mktime(created_date.timetuple())
    os.utime(download_path, (ctime, ctime))

sensei-ishikawa:
the ugly code worked. 
119195it [41:09:14,  1.24s/it] 2021-01-02 05:50:31 INFO     All photos have been downloaded!

i read the docs for time() and re-wrote the code accordingly. i am running it again. if you give me some pointers, i can write the cassette for functional testing. i renamed the variable ctime as epoch for clarity.

def set_utime(download_path, created_date):
    """Set date & time of the file"""
    try:
        epoch = time.mktime(created_date.timetuple())
    except OverflowError or ValueError:
        epoch = time.time()
    os.utime(download_path, (epoch, epoch))

AndreyNikiforov:
Folks, if we are dealing with datetimes without epoch, would it be reasonable to set epoch for these cases instead of catching overflow exception?

Here is a snipped from my unpublished code to explain the idea:

```
def datetime_cleansed(timestamp_ms: int) -> datetime.datetime:
    """
        Takes asset date as timestamp in ms and fallsback to 0
    """
    date_time = datetime.datetime.fromtimestamp(timestamp_ms / 1000, tz=datetime.timezone.utc)
    if date_time.year == datetime.MINYEAR and date_time.month == 1 and date_time.day == 1:
        # timestamp was built from time only
        return date_time.replace(year=1970)
    if date_time.year < 100:
        # missing epoch
        if date_time.year >= 70:
            return date_time.replace(year=1900+date_time.year)
        return date_time.replace(year=2000+date_time.year)
    return date_time
```

PR #289, opened 2021-01-15
Fix 286

menkej:
Fix the warning in the docker version.

This pull also includes an unrelated change to the test script by accident. Should be of no harm, trust me. ;-)

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -6,9 +6,14 @@ realpath() {
     [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
 }
 
+if [ $# -gt 0 ] ; then
+	numprocesses=$1
+else
+	numprocesses='auto'
+fi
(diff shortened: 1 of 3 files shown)


scripts/test
```diff
@@ -6,9 +6,14 @@ realpath() {
     [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
 }
 
+if [ $# -gt 0 ] ; then
```
AndreyNikiforov, line 9: That change does not seems to be related to timezone..

AndreyNikiforov (approved):

scripts/test
menkej: No, it is not. It's just a little addition I had in that script to limit the number of concurrent runs. I noticed when I created the pull request. Should be of no harm... ;-)

PR #133, opened 2019-04-03
Initial work towards HEIC to JPG conversion

chrisayoub:
This pull request relates to issue:
https://github.com/ndbroadbent/icloud_photos_downloader/issues/108

I have included code changes that use ImageMagick to convert downloaded HEIC images to JPG images. I have also updated the README to account for the new command line option.

Additionally, the README includes instructions on how to install ImageMagick on various platforms with HEIC support. So far, I have verified these steps on Ubuntu and Windows 10, and the Mac steps seem correct, but I am unable to formally verify this.

I could not figure out how to create a test case for this. Looking at the existing test cases, I am sure this is possible, but the VCR cassette tests were too complicated for me to manually modify to include an HEIC file (so conversion can be tested). Help would be appreciated with creating a test case for this.

Additionally, in running the existing test cases on my system (Ubuntu, with timezone GMT -6), I found that two of the test cases fail due to my system's timezone offset. I have included fixes for these specific testcases.

PR #322, opened 2021-04-22
HEIC to JPG image conversion

cdraf:
I was inspired by the [Initial work towards HEIC to JPG conversion](https://github.com/icloud-photos-downloader/icloud_photos_downloader/pull/133) and thought I could contribute. What I came up with is similar but does not require ImageMagick.

The initial PR, and subsequent commit (since squashed), triggered a git workflow that showed build errors for python3.9. It seems that there is no wheel but one is [coming](https://github.com/carsales/pyheif/pull/40).

Also the build machine needs libheif-dev installed to pass the check_quality and validate_build. I don't know if adding an additional package to the build/user machine falls in line with what this project wants but I think the build will pass if it is installed.


cdraf:
It looks like there is no wheel for python3.9 and pyheif but one is [coming](https://github.com/carsales/pyheif/pull/40).

Also it looks like the check_quality and validate_build need the **libheif-dev** package installed to build. I don't know if the extra package on the build/user machine falls in line with what this project wants but I think that will resolve the issue.

menkej:
Sorry for reacting so late, I'm really busy outside my open source dev work since a while (hope it will get better soon). Anyways:

Nice work! I could only have a quick look, so here are my initial thoughts: it adds another dependency to an external non python tool/lib, right? How do we make sure this still works on our currently supported platforms. There are users on Win/Mac and not only Linux relying on icloudpd. How could we handle this? It seems like you're sticking within docker at the moment. Have you tried other platforms?

cdraf:
> get some time in the next few weeks

that did not happen lol

I've changed base.py to import `heif_to_jpg` only if the `--convert-jpg` argument is passed which makes the dependency on pyheif (and Pillow) optional, end users will only need the additional dependencies if they want to do image conversion.

To handle the Windows case I built and installed `libheif` using vcpkg and copied the libs and dlls to the venv which allows the `pip install pyheif` to succeed. I've added instructions to the README.

I don't have a mac so I am not able to confirm if any of this works.

PR #329, opened 2021-05-07
Add option for a sender e-mail address

sgraaf:
This could prove useful when using a transactional e-mail service like Mailjet, where the SMTP username is a string of length 32 (and not an actual e-mail address).


sgraaf:
> 
> 
> Thanks for your pull requests. To be included it needs to be covered by unit tests. Could you please add testcasea for the new option?
> 
> Cheers, Jörg

I'll get on that!

PR #336, opened 2021-05-27
add new thumbnail flag

framirez90:
I added a new feature to only be able download the first photo of certain album as as it shown on the device.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ Options:
                                   Download live photos)
   --force-size                    Only download the requested size (default:
                                   download original if size is not available)
+  --thumbnail                     Only download the first photo as thumbnail 
+                                  of album as it is shown in the device 
+                                  (-a, --album option is required)
   --auto-delete                   Scans the "Recently Deleted" folder and
                                   deletes any files found in there. (If you
                                   restore the photo in iCloud, it will be
--- a/tests/test_download_photos.py
+++ b/tests/test_download_photos.py
@@ -1295,3 +1295,51 @@ class DownloadPhotoTestCase(TestCase):
                 photo_modified_time.strftime('%Y-%m-%d %H:%M:%S'))
 
             assert result.exit_code == 0
+
+    def test_get_thumbnail_photo_of_album(self):
+        base_dir = os.path.normpath(f"tests/fixtures/Photos/{inspect.stack()[0][3]}")
+        if os.path.exists(base_dir):
+            shutil.rmtree(base_dir)
+        os.makedirs(base_dir)
+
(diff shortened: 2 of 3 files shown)


menkej:
Thanks for your pull request! What is your usecase for this?

Could you please also add testcases for it?

framirez90:
Hi @menkej 

I want to retrieve only the photo as it shown on album cover on IOS device.

I added a new TC for it.

tests/test_download_photos.py
```diff
@@ -1295,3 +1295,51 @@ def test_download_chinese(self):
...
+                        setattr(self, "already_authenticated", True)
+
+                with mock.patch("icloudpd.constants.WAIT_SECONDS", 0):
+                    with mock.patch.object(
+                        PyiCloudService, "authenticate", new=mocked_authenticate
+                    ):
+                        runner = CliRunner(env={
+                            "CLIENT_ID": "DE309E26-942E-11E8-92F5-14109FE0B321"
+                        })
+                        result = runner.invoke(
```
menkej, line 1330: You are running the script but don't have any assertions/tests. I suggest checking for the pictures you are expecting to be downloaded. Additionally you should do a negative test, p.e. run without your new parameter (so that more pictures are downloaded) and check that the second picture that would come is not been downloaded when run with `--thumbnail`.

tests/test_download_photos.py
```diff
@@ -1295,3 +1295,51 @@ def test_download_chinese(self):
...
 
             assert result.exit_code == 0
+
+    def test_get_thumbnail_photo_of_album(self):
+        base_dir = os.path.normpath(f"tests/fixtures/Photos/{inspect.stack()[0][3]}")
+        if os.path.exists(base_dir):
+            shutil.rmtree(base_dir)
+        os.makedirs(base_dir)
+
+        with vcr.use_cassette("tests/vcr_cassettes/listing_photos.yml"):
```
menkej, line 1305: Best would be to add a new cassette for this test case. Creating a new cassette is pretty easy. Just use a new filename and run the test again. Make sure to remove personal data from the cassette before you check in.

README.md
```diff
@@ -52,6 +52,9 @@ Options:
                                   Download live photos)
   --force-size                    Only download the requested size (default:
                                   download original if size is not available)
+  --thumbnail                     Only download the first photo as thumbnail 
+                                  of album as it is shown in the device 
+                                  (-a, --album option is required)
```
menkej, line 57: why is `-a` required?

menkej (changes requested):
Please have another look at the test case and also add your change to the CHANGELOG. Thanks!

PR #459, opened 2022-08-15
GitHub Action to run tox

cclauss:
Test results: https://github.com/cclauss/icloud_photos_downloader/actions

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,0 +1,18 @@
+name: tox
+on: [push, pull_request]
+jobs:
+  tox:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]  # [macos-latest, ubuntu-latest, windows-latest]
+        python: ['3.x']  # ['3.7', '3.8', '3.9', '3.10', '3.11-dev', 'pypy-3.9']
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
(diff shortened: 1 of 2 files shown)


AndreyNikiforov:
@cclauss what value does `tox` give us in addition to existing quality checks? (I am not very familiar with it)

.github/workflows/tox.yml
```diff
@@ -0,0 +1,18 @@
+name: tox
+on: [push, pull_request]
+jobs:
+  tox:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]  # [macos-latest, ubuntu-latest, windows-latest]
+        python: ['3.x']  # ['3.7', '3.8', '3.9', '3.10', '3.11-dev', 'pypy-3.9']
```
AndreyNikiforov, line 9: We currently support 3.6-3.9 (and have issues with 3.10+). Should we limit python versions to only that subset?

AndreyNikiforov (changes requested):

.github/workflows/tox.yml
cclauss: GitHub Actions ubuntu-latest no longer supports Python 3.6 because the Python Team no longer supports it.
* https://devguide.python.org/versions

I have set the tests to run on 3.7-3.9

cclauss:
I did not add `tox.ini` -- It was already there.  If the existing Quality Checks run the same tests as `tox.ini` the we can close this pull request and delete `tox.ini`.

AndreyNikiforov:
> If the existing Quality Checks run the same tests as `tox.ini` then we can close this pull request and delete `tox.ini`.

Can you check if they are indeed the same and if they are remove `tox` traces, please?

PR #431, opened 2022-05-10
Delete photo/video after download successful

MatrixDai:
Added a new command line option `--delete-after-download` to allow user delete the photo right after download complete. The deleted items will moved to "Recent Deleted" system album.

The implementation reference from [this gist](https://gist.github.com/Tatsh/0b5a10a2d25f6c8466290401d2026055).

Related discussions:
#368 #314 #124 #332

PR #525, opened 2023-01-12
support to delete downloaded photos

ryda20:
Hi there!

Thank you so much for spending time to write this tool!!

In my case, I need to delete all photos from iCloud after downloaded to save space for next sync from my iPhone (free plan only have 5GG), so, after searching and testing, I already added "--auto-delete-downloaded" flag.

I think, it is not only my needed case, someone else will need. So, Please kindly take a look if you want to add this feature!

Thank you so much!

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Options:
                                   deletes any files found in there. (If you
                                   restore the photo in iCloud, it will be
                                   downloaded again.)
+  --auto-delete-downloaded        Auto delete downloaded photo from icloud.
   --only-print-filenames          Only prints the filenames of all files that
                                   will be downloaded (not including files that
                                   are already downloaded.)(Does not download
--- a/icloudpd/base.py
+++ b/icloudpd/base.py
@@ -211,6 +211,7 @@ def main(
         skip_live_photos,
         force_size,
         auto_delete,
+        auto_delete_downloaded,
         only_print_filenames,
         folder_structure,
         set_exif_datetime,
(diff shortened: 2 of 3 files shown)


AndreyNikiforov:
@ryda20 thanks a lot for contribution. Quick question: what happens if user specifies existing `--auto-delete` option and new `-auto-delete-downloaded` together? And if running multiple times with both options? I did not analyze all possible outcomes, but it seems that suicide is possible (loosing all assets locally and in the cloud). If that is correct, how can we help users avoid that outcome?

README.md
```diff
@@ -56,6 +56,7 @@ Options:
                                   deletes any files found in there. (If you
                                   restore the photo in iCloud, it will be
                                   downloaded again.)
+  --auto-delete-downloaded        Auto delete downloaded photo from icloud.
```
AndreyNikiforov, line 59: What can we do to help users avoid running `--auto-delete-downloaded` and `--auto-delete` together or consequently (which will delete all photos locally and in the cloud)?

icloudpd/base.py
```diff
@@ -543,7 +554,37 @@ def download_photo(counter, photo):
                         download.download_media(
                             icloud, photo, lp_download_path, lp_size
                         )
-
+    
+    photos_delete_list = []
+    def auto_delete_downloaded_photos(photo, end=False):
```
AndreyNikiforov, line 559: Can we have a test for this new functionality, please?

AndreyNikiforov (changes requested):
Please rebase to get tests working, add tests for new functionality, and suggest idea to prevent total loss of photo assets (locally and in the cloud).

AndreyNikiforov:
There is similar feature in #431 . What are the differences between the two?

Issue #179, opened 2020-10-21
Remove pyicloud_ipd fork


AndreyNikiforov:
@ndbroadbent what is in pyicloud_ipd fork that cannot be merged into upstream? Using upstream will hopefully reduce support scope.

ndbroadbent:
Yes I totally agree, I think it's definitely time to do this! That would be great! There's a ton of minor changes and fixes, and I looked into this a few weeks ago before giving up. But I think it might not be too hard to get everything removed and updated.

These are all the changes in the current version that is being used:

```
* 1d00ee5 - (HEAD -> pyicloud-ipd) Added build script (3 weeks ago) <Nathan Broadbent>
* 8f5bab0 - Fix release script (3 weeks ago) <Nathan Broadbent>
* 7899300 - Bump version (3 weeks ago) <Nathan Broadbent>
* 1f0d5c7 - (origin/pyicloud-ipd) Apply change from https://github.com/ndbroadbent/pyicloud/pull/4 (3 weeks ago) <Nathan Broadbent>
* 37e25bd - Bump version to 0.10.0 (1 year, 5 months ago) <Nathan Broadbent>
* a1a9f21 - Fix crash when album titles contain unicode characters (e.g. umlauts) (1 year, 5 months ago) <Nathan Broadbent>
* 5620b74 - Bump version to 0.9.9 (1 year, 10 months ago) <Nathan Broadbent>
* e3197df - Fix import in icloud cli tool (1 year, 10 months ago) <Nathan Broadbent>
* c16e67b - Bump version to 0.9.8 (1 year, 11 months ago) <Nathan Broadbent>
* 9e146ba - When filenameEnc is missing, build a filename from the first 12 characters of the sanitized fingerprint and the correct file extension (1 year, 11 months ago) <Nathan Broadbent>
* e8bc1cd - (tag: 0.9.7) Bump version to 0.9.7 (2 years ago) <Nathan Broadbent>
* 3ee4ea3 - Added an optional exception_handler to the PhotoAlbum photos iterator, so that we can catch and retry after authentication errors (2 years ago) <Nathan Broadbent>
* 5efd010 - Added install script to make development easier (2 years ago) <Nathan Broadbent>
* 28fd82a - Fix filenames for live photos (2 years, 2 months ago) <Nathan Broadbent>
* f741d8b - Added some build & release scripts (2 years, 2 months ago) <Nathan Broadbent>
[...]

PR #528, opened 2023-01-17
update to pyicloud-1.0.0 to support 2fa verification

qzchenwl:
fixes https://github.com/boredazfcuk/docker-icloudpd/issues/227


AndreyNikiforov:
@qzchenwl looks like you are removing dependency on pyicloud fork - a long standing request #179 . That is awesome and thank you very much for the effort.

How can we validate that such massive change does not break anything? 

As a small step, may be we can make existing tests working? I recently fixed pipeline and all tests on master are green. Can you rebase/pull forward from the master to take advantage of the recent fixes in quality checks, please?

qzchenwl:
> @qzchenwl looks like you are removing dependency on pyicloud fork - a long standing request #179 . That is awesome and thank you very much for the effort.
> 
> How can we validate that such massive change does not break anything?
> 
> As a small step, may be we can make existing tests working? I recently fixed pipeline and all tests on master are green. Can you rebase/pull forward from the master to take advantage of the recent fixes in quality checks, please?

I tried to make existing tests working #536. I've updated the Python code part of the unit test, but I failed to update the http request mocking because I am not familiar with pytest and vcr.

The reason we need to update http request mocking is that pyicloud-1.0.0 use `https://idmsa.apple.com/appleauth/auth/signin` to authenticate while current test case only mocks  `https://setup.icloud.com/setup/ws/1/login`.

What's the best practice to mock new http request (maybe record and save real requests) without using a real Apple ID?

qzchenwl:
@AndreyNikiforov I don't think http mocking style unit test is wrong way to test, because it is not the responsibility 
 of icloud_photos_downloader to abstracting the implementation detail (rest api), it is the responsibility of pyicloud. The unit tests should mock the public api of pyicloud.

menkej:
> What's the best practice to mock new http request (maybe record and save real requests) without using a real Apple ID?

This is the way I'd go. Just record a new request using a valid Apple ID and then remove the Credentials from the recording before checking it in.

AndreyNikiforov:
> @AndreyNikiforov I don't think http mocking style unit test is wrong way to test, because it is not the responsibility of icloud_photos_downloader to abstracting the implementation detail (rest api), it is the responsibility of pyicloud. The unit tests should mock the public api of pyicloud.

@qzchenwl I agree that having unit test for our logic and give responsibility for testing http parsing to pyicloud is a good approach. 

We can still benefit for e2e tests to confirm that our product works as intended. I see existing tests with http request mocking kinda e2e (but without real connection). Even though they are not making real connection they are still a good protection against regressions and we better keep them relevant.

Issue #146, opened 2020-04-01
Help creating an EXE


wes1993:
Hello!!
I love this project and i want to save in my pc.
can you explain to me how can i build and EXE so i don't need to install everytime python etc?
Thanks a lot
Stefano

metahnetoutage:
Hello,

To save you time, it is cheaper for you to learn how to connect to iCloud
via JavaScript api provided by Apple.

You can do authentication and not waste time on this project. You will
still be Abel to do so much directly with Apple stuff. Why? Because this
project does bad hack job.

It is not supported anymore, it is for older versions of iPhone until
around 5/6 maybe 7.

Spend quality 40 hours learning how to connect via JavaScript and do
authentication and spend day $100 on Apple license if you want to do more.

That is time and money well spent.

You expectations for this project are at 9, when it is at 5 level at best.

I am a developer.

Dima

wes1993:
Thanks a lot for your reply.
I’m not a developer so I don’t know how to use Apple api... also is for personal use so I 100€ day is so much expensive for me... :(
Do you know different software for downloading photos from iCloud?
The Apple iCloud software and iCloud.com won’t work very well... :(

metahnetoutage:
I was going to do my own...This is the closest I found.

To help you out your comment should be:
Open up a terminal and navigate to directory where you’d aced this python
file.
Name of program -d /pathtoDirectoryWhereyousave

The system will prompt you for user name
And password. Enter both

So suppose you saved the file at mycooldir.

You would cd there. Then type name of application -d ./pathtodestination

If you check docs you will see this, now, make sure you have -d before your
directory. I replied once in a threat you can see my answer.

Let me know if you need any extra help.
Session is 5 mins so do not download a lot.

Dima

Quentame:
@metahnetoutage

Do you have a link to the iCloud JS API ?

I found this.

Thanks.

metahnetoutage:
I can give you yes.
https://developer.apple.com/documentation/cloudkitjs

Take your Time and do not rush. It’s is not a one two three.

RhetTbull:
This might help: [auto-py-to-exe](https://github.com/brentvollebregt/auto-py-to-exe)  It can convert the python script into a stand-alone .exe

wes1993:
Thanks a lot to everyone for the help!!! :-D
sorry for my late reply but i'm in Italy...
[...]

PR #529, opened 2023-01-17
package icloudpd as one executable

AndreyNikiforov:
A lot of support questions come from people who struggle to configure pre-requisites (python) correctly. Packaging python and icloudpd into one binary with no pre-requisites will help a lot.

As a side effect, python 3.6 had to be deprecated as it is not supported by the tool used for packaging (pyinstaller)

Mentioned in #146


menkej:
Hi Andrey! You've been busy! I'll try to have a look at it soon, but I'm currently buried in some private and work related challenges, so I'll probably not be able to have a look at it within the next two weeks. I'll try.

PR #537, opened 2023-01-20
Release v1.7.3

AndreyNikiforov:
Update version and changelog in preparation for v1.7.3 release

### DIFF
--- a/.github/workflows/quality-checks.yml
+++ b/.github/workflows/quality-checks.yml
@@ -5,7 +5,8 @@ name: Quality Checks
 
 on:
   push:
-    # branches: [ master ]
+    branches:
+      - '**'
   pull_request:
     # branches: [ master ]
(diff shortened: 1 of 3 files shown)


ashleycawley:
Hey Andrey, thank you for all you do, I think you might have put the wrong year by mistake in your change log, it reads:
1.7.3 (2022-01-20)
But seeing as it was updated on the 20th Jan **_2023_** I'm guessing it should be 2023?
All the best.

AndreyNikiforov:
> Hey Andrey, thank you for all you do, I think you might have put the wrong year by mistake in your change log, it reads: 1.7.3 (2022-01-20) But seeing as it was updated on the 20th Jan **_2023_** I'm guessing it should be 2023? All the best.

Indeed... fixed. Thanks for bringing it to my attention!

PR #610, opened 2023-03-21
Skip download if file exists without checking for file size (in bytes)

ahti123:
Docker online-only files report 0 bytes size on newer macOS versions. As I keep "local" iCloud photos copy on in Docker it is useful to ignore file size on download.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ Options:
                                   will be downloaded (not including files that
                                   are already downloaded.)(Does not download
                                   or delete any files.)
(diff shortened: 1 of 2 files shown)


AndreyNikiforov:
@ahti123 can you think of a way to test your code (with automated testing)?

AndreyNikiforov:
@ahti123 you can run locally:
- `scripts/install_deps`
- `scripts/lint`
- `scripts/test`

That should save you some time waiting for github actions.

ahti123:
Ok, will lint and test

PR #613, opened 2023-03-28
Skip download if file exists without checking for file size (in bytes) 

ahti123:
Added a cmdline switch to skip re-downloading if the size (in bytes) does not match.
Docker online-only files report 0 bytes size on newer macOS versions. As I keep "local" iCloud photos copy on in Docker it is useful to ignore file size on download.
Linter has no complaints.
Tests added for the new --ignore-local-filesize switch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ Options:
                                   will be downloaded (not including files that
                                   are already downloaded.)(Does not download
                                   or delete any files.)
+  --ignore-local-filesize         Skipping download if file(name) exists
+                                  locally regardless of the difference in size.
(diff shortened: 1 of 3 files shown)


AndreyNikiforov:
> ...Docker online-only files...

Based on comments in source code, you meant "Dropbox", not "Docker", correct?

Can you add a line to CHANGELOG to announce the change in the next release, please?

ahti123:
Yeah, type - Dropbox 0-bytes files.
Devel process-wise: would be good to create a feature request issue. And I discovered a bug, too :) Withdrawing the pull request.

PR #620, opened 2023-04-10
Added option to download X oldest items, not only X recent ones.

Santa77:
Added option to download X oldest items, not only X recent ones.


sarambasich:
I, too, would like to see this feature implemented. Is all that's missing to merge this test coverage?

Issue #614, opened 2023-04-01
`delete-after-download` deletes photos from `iCloud` that haven't been successfully downloaded


jpavlav:
## Overview
When specifying the `delete-after-download` flag on CLI the media item is deleted from `iCloud` whether or not the download succeeds. This is, I'd wager, something that folks will be surprised by (I definitely was).

## Steps to Reproduce
[NOTE]: # ( Provide a simple set of steps to reproduce this bug. )
1. Execute any download command that includes the `delete-after-download` flag:
```
icloudpd --directory . --username <user> --delete-after-download --recent 100 --folder-structure <photo_dir>
```
2. I think the check to reproduce this, you probably need to short circuit the [download_media](https://github.com/icloud-photos-downloader/icloud_photos_downloader/blob/18efdfe90e222fac3ef5a33072bc15fdd072dde4/icloudpd/download.py#L35) function to skip to `else` condition.
3. Observe the ouput, the file is not downloaded, but it _is_ deleted from `iCloud`:
```
Error downloading RPReplay_Final1651790030.mp4, retrying after 5 seconds...
Error downloading RPReplay_Final1651790030.mp4, retrying after 10 seconds...
Error downloading RPReplay_Final1651790030.mp4, retrying after 15 seconds...
Error downloading RPReplay_Final1651790030.mp4, retrying after 20 seconds...
Error downloading RPReplay_Final1651790030.mp4, retrying after 25 seconds...
Could not download RPReplay_Final1651790030.mp4! Please try again later.
Downloading/RPReplay_Final1651790030.mp4:  68%|6| 68/102023-04-01 01:00:50 INFO     Deleting RPReplay_Final1651790030.mp4
```

## Expected Behavior
It makes sense, that if a download fails, that the file in `iCloud` should _not_ be deleted. It puts the user in a position where they now have to negotiate restoring a photo/video, instead of just trying to download it again in the normal fashion.

## Actual Behavior
The file is not downloaded to disk and _is_ deleted from `iCloud`. In my case, I'm going to have to now figure out which files I don't have locally and attempt to restore all of them (hundreds, potentially over a thousand files) so that I can download them again.

## Context
I started a branch to work on this, but the issue is that the result of the download is nested so far away from the code that governs deleting after download, it would either require passing in the `delete-after-download` flag to the `download` function (which might work) and make the adjustment there. 

[...]

PR #625, opened 2023-04-19
Only delete files that were successfully downloaded

jfsimoneau:


AndreyNikiforov:
@jfsimoneau do you have an idea what tests can a good indicator that new functionality is fixing the issue and working correctly? That would be very helpful for maintaining the app going forward.

jfsimoneau:
I added a test with a failed download. This should fix #614.

tests/test_download_photos.py
```diff
@@ -1506,6 +1506,61 @@ def test_download_after_delete(self):
...
+                self.assertIn("DEBUG    Looking up all photos from album All Photos...", self._caplog.text)
+                self.assertIn(
+                    f"INFO     Downloading the first original photo to {base_dir} ...",
+                    self._caplog.text,
+                )
+                self.assertIn(
+                    f"INFO     Downloading {os.path.join(base_dir, os.path.normpath('2018/07/31/IMG_7409.JPG'))}",
+                    self._caplog.text,
+                )
+                self.assertNotIn(
```
AndreyNikiforov, line 1552: Is there a way to confirm skipping deletion call without/in addition to relying on logs? It would be beneficial to separate logic check from logging check, so we can change one without breaking tests for the other.

Note the most test still combine log and state/file/api checks and I hope that can be changes over time too.

jfsimoneau: I wanted to mock the internal `delete_photo()` function to see if it was called, but it seems it's not accessible. The best way I found to make that tests is to craft a cassette to simulate a bad request. Because of that bad request, the delete is never called so the cassette is complete.

PR #645, opened 2023-06-24
Auto delete date fix testing

steffenmartin:
@AndreyNikiforov, here's my first attempt of a comprehensive fix for this bug. I hope this is in adherence to all contributing guidelines.
Since test _test_autodelete_photos_ inside of _tests/test_autodelete_photos.py_ was already performed incorrectly it is somewhat difficult to perform a before-after comparison. If that's needed I suggest to first apply the follow patch to it.
image
This will make it fail in its current state. And once this PR has been merged it will pass again.


AndreyNikiforov:
@steffenmartin some tests are failing - pls fix. the PR should have all tests green. If anybody decides to validate fix for the test before the fix for the code, they will have to change some files locally IMO.

steffenmartin:
> @steffenmartin some tests are failing - pls fix. the PR should have all tests green. If anybody decides to validate fix for the test before the fix for the code, they will have to change some files locally IMO.

Ok, ready for a re-run.

PR #661, opened 2023-07-12
offer to save user password if it is entered interactively with the icloudpd command

mzhou:


AndreyNikiforov:
Thanks for new feature. A number of questions since this feature was not discussed before: What will be new experience for users who do not use keyring to store passwords? What happens when keyring is not available?

Issue #690, opened 2023-09-26
Listing libraries returns INTERNAL_ERROR


gklka:
## Overview

First of all, thank you for the shared libraries support! I was really looking forward for it.

Using `--list-libraries` switch results the following:

## Steps to Reproduce

```
 $ sudo /opt/icloudpd/bin/icloudpd --directory /mnt/disk7/iCloud \
           --username **** \
           --password **** \
           --list-libraries
2023-09-26 07:49:43 DEBUG    Authenticating...
2023-09-26 07:49:47 ERROR    INTERNAL_ERROR (INTERNAL_ERROR)
Traceback (most recent call last):
  File "/opt/icloudpd/bin/icloudpd", line 8, in <module>
    sys.exit(main())
  File "/opt/icloudpd/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/opt/icloudpd/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/opt/icloudpd/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/icloudpd/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/opt/icloudpd/lib/python3.10/site-packages/icloudpd/base.py", line 313, in main
    core(
  File "/opt/icloudpd/lib/python3.10/site-packages/icloudpd/base.py", line 767, in core
    libraries_dict = icloud.photos.libraries
  File "/opt/icloudpd/lib/python3.10/site-packages/pyicloud_ipd/services/photos.py", line 277, in libraries
    libraries[zone_name] = PhotoLibrary(
  File "/opt/icloudpd/lib/python3.10/site-packages/pyicloud_ipd/services/photos.py", line 154, in __init__
    request = self.service.session.post(
  File "/opt/icloudpd/lib/python3.10/site-packages/requests/sessions.py", line 637, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
  File "/opt/icloudpd/lib/python3.10/site-packages/pyicloud_ipd/base.py", line 105, in request
    self._raise_error(code, reason)
  File "/opt/icloudpd/lib/python3.10/site-packages/pyicloud_ipd/base.py", line 127, in _raise_error
    raise api_error
[...]

PR #692, opened 2023-09-27
fixing #690 internal_error

fver:


fver:
Changed URL in libraries() call in pyicloud_ipd/services/photos.py:261 to url = ('%s/zones/list' %
adapted tests and VCR files
Nu further validation since I was unable to reproduce the original problem, but at least one person confirmed it was fixed.

PR #712, opened 2023-11-08
Upgrade python to 3.12 and upgrade workflow runner

harryzcy:


harryzcy:
The failing test is related to https://github.com/PythonCharmers/python-future/issues/625

src/pyicloud_ipd/services/photos.py
```diff
@@ -10,7 +10,12 @@
 
 import pytz
 
-from future.moves.urllib.parse import urlencode
+try:
+    # Python 3
+    from urllib.parse import urlencode
+except ImportError:
+    # Python 2
```
AndreyNikiforov, line 17: do we need that at all? we do not support python 2 any more...

.github/dependabot.yml
```diff
@@ -10,3 +10,13 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 5
+  - package-ecosystem: "docker"
```
AndreyNikiforov, line 13: what value are we expecting from dependabot? In the past it was just too noisy. I was considering removing it altogether...

AndreyNikiforov:
> The failing test is related to [PythonCharmers/python-future#625](https://github.com/PythonCharmers/python-future/issues/625). It's only imported once, we can easily remove it.

You can run test locally. If you are using dev container, then switching python versions should be easy and you can validate tests in all supported python versions -- that will be faster than waiting for github actions approved.

.github/dependabot.yml
harryzcy: > In the past it was just too noisy. I was considering removing it altogether...

Ok, sure. I can revert the changes then

.github/workflows/codeql-analysis.yml
```diff
@@ -24,7 +24,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-20.04
```
AndreyNikiforov, line 27: Can we use exact version, pls? That should prevent surprises

PR #778, opened 2024-01-28
Add AUR link to README

undefined-landmark:
Hi,

Thanks for the nice tool! Since a package was not available on the AUR yet I created a simple PKGBUILD that sets up the icloudpd binary. This should make using and updating the tool a bit easier for Arch users. Because this provides another way to install the package I added this option to the README. If there is any additional documentation you'd like me to write I'd be happy to do so, just let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ We aim to release new versions once a week (Friday), if there is something worth
 
 There are three ways to run `icloudpd`:
 1. Download executable for your platform from the Github [Releases](https://github.com/icloud-photos-downloader/icloud_photos_downloader/releases) and run it
-1. Use package manager to install, update, and, in some cases, run ([Docker](README_DOCKER.md), [PyPI](README_PYPI.md), Experimental: [Npm](README_NPM.md))
+1. Use package manager to install, update, and, in some cases, run ([Docker](README_DOCKER.md), [PyPI](README_PYPI.md), [AUR](https://aur.archlinux.org/packages/icloudpd-bin), Experimental: [Npm](README_NPM.md))
 1. Build and run from the source
 
 ## Features


AndreyNikiforov:
The idea to support native package support for linux is great! 

For Arch linux it is pacman, so end result of supporting pacman should be that end user should be able to install icloudpd using, say, `pacman -S icloudpd`. Are we getting that user experience by adding AUR? (I not familiar with Arch/pacman/AUR).

What are the capabilities for permission control in pacman registry? For supported registries/repositories/etc we try to keep accounts of individual members/admins organized into org/group, so that access is not locked to one individual.

undefined-landmark:
AUR packages are first compiled, manually or by an [AUR helper](https://wiki.archlinux.org/title/AUR_helpers), and then installed through pacman. A basic description of AUR is best found on the wiki [wiki](https://wiki.archlinux.org/title/Arch_User_Repository). Via the yay AUR helper an install would only be a simple `yay -S icloudpd-bin`.

AUR allows for co-maintainers. So I could add additional people if available. If for some reason I would stop maintaining the package suddenly the package can be flagged though an 'Orphan Request'. If I don't respond someone else will be able to take over maintenance. Exact details can be found on the [wiki](https://wiki.archlinux.org/title/AUR_submission_guidelines#Requests) again :)

AndreyNikiforov:
I looked at the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=icloudpd-bin) -- it is good that you are using existing released binaries instead recompiling. How can AUR/PKGBUILD support multiple architectures? icloudpd and archlinux both support amd64, arm32, and arm64 IIUC.

undefined-landmark:
There are builtin PKGBUILD options to allow for installation on different architectures. Basically comes down to adding extra sources and checksums for each architecture. I can take a look and update the PKGBUILD accordingly.

AndreyNikiforov:
> There are builtin PKGBUILD options to allow for installation on different architectures. Basically comes down to adding extra sources and checksums for each architecture. I can take a look and update the PKGBUILD accordingly.

Awesome! May I also suggest renaming package to `icloudpd` (remove `-bin`). To end user it is irrelevant information -- they just install icloudpd and the fact that it uses precompiled binaries instead of building from the source should/can be hidden.

undefined-landmark:
The bin suffix is an AUR submission rule. It indicates that the package uses prebuilt binaries. No suffix would indicate the package is built from source. So with the current PKGBUILD I can't just remove it unfortunately.

Making a PKGBUILD that builds from source is possible of course. Albeit a bit more effort.

Reference: 
[AUR guidelines](https://wiki.archlinux.org/title/AUR_submission_guidelines)
> Packages that use prebuilt deliverables, when the sources are available, must use the -bin suffix. An exception to this is with Java. The AUR should not contain the binary tarball created by makepkg, nor should it contain the filelist.

undefined-landmark:
The [PKGBUILD](https://aur.archlinux.org/packages/icloudpd-bin) now supports armv7h, arm64 and i686. I don't have those architectures around to test but, assuming I matched the architectures and bins correctly, they should function.

undefined-landmark:
Thanks for the quick merge. Just saw there's another set of download links at the start of the README. Bit inattentive on my part. Something I noticed before is that the links point to README's of each download option, instead of direct links. Is there any need for such a README for the AUR option?

Can fix both of course. But already great that a direct link is visible in the README now.

Issue #466, opened 2022-08-27
QUESTION: Is it possible to download files based on date?


attzonko:
## Summary
I see the setting where you can specify to download the N most recent photos. Is it possible to download all the most recent photos taken over the last 30 days, or since a specific date?

## Context
I would like to automate this to sync my icloud photos periodically grabbing all the photos since the last sync point. I realize it doesn't re-download photos which are already present, however, my use case is that the pictures get moved/renamed from the raw download area to a different area.

menkej:
Currently this is not possible. Someone feel free to add this feature and submit a pull request.

raleighlittles:
@attzonko What exactly do you mean by this?

> I realize it doesn't re-download photos which are already present

How do you know the tool doesn't re-download photos/videos that are already downloaded?

Jonnypleasehelp:
Yes I have early photos from 2004

Brett-StrayThought:
I really want this feature request too. I want to copy images equal to or later than a date specified. Allows me to sync my media library (which is processed, renamed and replicated to other places) periodically.

AndreyNikiforov:
@attzonko @Brett-StrayThought IIUC you download photos, then reorg/move them locally, and at the next sync want to download newly _added_ photos. If you uploaded to iCloud.com some photos taken years ago (`date_created` is old), they are still downloaded during next `icloudpd` run.

How do you coordinate `icloudpd` runs and reorg of your libraries? I suspect it has to be very tight orchestration between the two, which most likely means manual runs.

gfody:
issue #660 could be way to support this workflow, assuming the internal state of a Copy command contained all the information it needed to discern between remote and local moves & renames

AndreyNikiforov:
1.29.0 has `skip-created-before` and `--skip-created-after` params

PR #779, opened 2024-01-28
Add command line arguments to filter by date.

itwasabhi:
With the new arguments --date-before and --date-after, user can choose to only download items that are created either before, after, or before and after the specified dates. Dates are assumed to be specified in the local timezone.

This should address issue #466, something I have also wanted for a while! 

Comments / suggestions welcome.

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -233,6 +233,16 @@ CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"]}
     type=click.Choice(["com", "cn"]),
     default="com",
 )
+@click.option("--date-before",
+              help="Latest date with which to filter which photo/videos will be downloaded, "
+              "specified in the format YYYY-MM-DD.",
+              default=None,
+              )
+@click.option("--date-after",
+              help="Earliest date with which to filter which photo/videos will be downloaded, "
+              "specified in the format YYYY-MM-DD.",
+              default=None,
+              )
 @click.option("--watch-with-interval",
               help="Run downloading in a infinite cycle, waiting specified seconds between runs",
               type=click.IntRange(1),
@@ -281,6 +291,8 @@ def main(
         threads_num: int,    # pylint: disable=W0613
         delete_after_download: bool,
         domain: str,
+        date_before: str,
+        date_after: str,
         watch_with_interval: Optional[int],
         dry_run: bool
 ):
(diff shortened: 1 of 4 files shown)


src/icloudpd/base.py
```diff
@@ -233,6 +233,16 @@
     type=click.Choice(["com", "cn"]),
     default="com",
 )
+@click.option("--date-before",
```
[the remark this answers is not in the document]
wasabhi93: Agreed, `created-before`  and `created-after` is much more direct and readable! updating.

wasabhi93:
@AndreyNikiforov should be ready for re-review

tests/test_download_photos.py
```diff
@@ -2293,3 +2293,99 @@ def raise_response_error(a0_, a1_, a2_):
...
+            self.assertIn(
+                "DEBUG    Looking up all photos from album All Photos...", self._caplog.text)
+            self.assertIn(
+                f"DEBUG    Skipping IMG_7409.JPG, date is after", self._caplog.text
+            )
+
+        files_in_result = glob.glob(os.path.join(
+            data_dir, "**/*.*"), recursive=True)
+
+        assert sum(1 for _ in files_in_result) == 0, "No files should have been downloaded."
```
AndreyNikiforov, line 2345: pls adjust the test to have a set of files to download with different created dates and then "cut" that set with your new parameter. Confirm that only desired portion of the set is downloaded. That should confirm that criteria filtering is correct and test itself is not broken (as it downloads some files).

There should be tests that validate filters in this manner.

AndreyNikiforov (changes requested):

src/icloudpd/base.py
```diff
@@ -418,6 +452,18 @@ def download_photo_(counter: Counter, photo: PhotoAsset) -> bool:
                     photo.created)
                 created_date = photo.created
 
+            if created_before and created_date > created_before:
```
AndreyNikiforov, line 455: These are the conditions we need to validate with a test. Parameter absence is tested, but comparison is not completely validated IMO - need to check we indeed leave some assets and filter out others (assets created on the same date as parameter is the case that should be present in tests as well). In e2e tests that will be translated into checking that some files are downloaded and some are not.

mrsilver76:
@itwasabhi Is there any chance you can do the changes requested? It's a useful feature so thank you for developing it.

itwasabhi:
@mrsilver76 Sure, please give me a week or so to sync to head and make the requested changes!

fliespl:
@itwasabhi do you think you will get back to it somewhere in the future :) ? This would a game-changer for date based deletion.

mrsilver76:
@itwasabhi Is there any chance you can make the changes requested by the code review?

I ask because I'm trying to create an automated process to download photos from iCloud and then group and move them elsewhere. As I don't keep the files in their original download location, I need some way to tell icloudpd not to download photos any earlier than the date the process was last run.   

Thank you!

AndreyNikiforov (changes requested):
pls adjust the test to have a set of files to download with different created dates and then "cut" that set with your new parameter. Confirm that only desired portion of the set is downloaded. That should confirm that criteria filtering is correct and test itself is not broken (as it downloads some files).

There should be tests that validate filters in this manner.

PR #915, opened 2024-07-17
Add captions and keywords as a photo field

adambenali:
iOS 14 introduced captions to Photos. This PR adds the caption field to the list of requested fields.

### DIFF
--- a/src/pyicloud_ipd/services/photos.py
+++ b/src/pyicloud_ipd/services/photos.py
@@ -501,7 +501,7 @@ class PhotoAlbum(object):
                 u'addedDate', u'isFavorite', u'isHidden', u'orientation',
                 u'duration', u'assetSubtype', u'assetSubtypeV2',
                 u'assetHDRType', u'burstFlags', u'burstFlagsExt', u'burstId',
-                u'captionEnc', u'locationEnc', u'locationV2Enc',
+                u'captionEnc', u'extendedDescEnc', u'locationEnc', u'locationV2Enc',
                 u'locationLatitude', u'locationLongitude', u'adjustmentType',
                 u'timeZoneOffset', u'vidComplDurValue', u'vidComplDurScale',
                 u'vidComplDispValue', u'vidComplDispScale',
--- a/src/pyicloud_ipd/services/photos.py
+++ b/src/pyicloud_ipd/services/photos.py
@@ -501,7 +501,7 @@ class PhotoAlbum(object):
                 u'addedDate', u'isFavorite', u'isHidden', u'orientation',
                 u'duration', u'assetSubtype', u'assetSubtypeV2',
                 u'assetHDRType', u'burstFlags', u'burstFlagsExt', u'burstId',
-                u'captionEnc', u'extendedDescEnc', u'locationEnc', u'locationV2Enc',
+                u'captionEnc', u'extendedDescEnc', u'keywordsEnc', u'locationEnc', u'locationV2Enc',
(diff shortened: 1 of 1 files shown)


AndreyNikiforov:
> iOS 14 introduced captions to Photos. This PR adds the caption field to the list of requested fields.

What problem does your change solve?

adambenali:
> > iOS 14 introduced captions to Photos. This PR adds the caption field to the list of requested fields.
> 
> What problem does your change solve?

@AndreyNikiforov This change makes it possible to read the keywords / caption attached to a picture from the `photo._asset_record` field. 

One example use-case is that I can decide to embed this information into the EXIF of the image after I download it.

AndreyNikiforov:
> @AndreyNikiforov This change makes it possible to read the keywords / caption attached to a picture from the `photo._asset_record` field.
> 
> One example use-case is that I can decide to embed this information into the EXIF of the image after I download it.

Ok, so you would like user to be able to enter captions in iCloud.com and would like downloaded images to have these captions in EXIF. I assume that user can edit captions in iCloud.com, there is a standard field in EXIF for captions, and Apple does not add these captions into EXIF of the asset automatically (did you check all of these conditions?).

Then your PR should have functionality to request extra field from iCloud service and update EXIF in images with it. Functionality should work only when user requested it (==flag parameter). There should be tests confirming that you new behavior works, see tests for EXIF dates.

adambenali:
> Then your PR should have functionality to request extra field from iCloud service
This is done.

> and update EXIF in images with it
This is out of the scope of this PR. It's up to the app using icloudpd to make use of this information. The point of this PR is to have this caption information be accessible as part of the `PhotoAsset` object.

All in all, this PR is only updating the long list of requested fields, nothing else. I don't think it requires that many extra steps as you're suggesting.

AndreyNikiforov:
> > and update EXIF in images with it
> 
> This is out of the scope of this PR. It's up to the app using icloudpd to make use of this information. The point of this PR is to have this caption information be accessible as part of the `PhotoAsset` object.
> 

`icloudpd` is not a library that some other app is using, instead it is an application that provides functionality to end users. You need to provide all other functionality that you envisioned. I suggest posting an issue with what you are trying to achieve, so your idea can be discussed in detail before spend time coding.

adambenali:
> > > and update EXIF in images with it
> > 
> > 
> > This is out of the scope of this PR. It's up to the app using icloudpd to make use of this information. The point of this PR is to have this caption information be accessible as part of the `PhotoAsset` object.
> 
> `icloudpd` is not a library that some other app is using, instead it is an application that provides functionality to end users. You need to provide all other functionality that you envisioned. I suggest posting an issue with what you are trying to achieve, so your idea can be discussed in detail before spend time coding.

Question: How is `captionEnc` (which is a requested field that's already there in the list) currently being used? 

Now the exact same thing applies to `extendedDescEnc` and `keywordsEnc`.

AndreyNikiforov:
> Question: How is `captionEnc` (which is a requested field that's already there in the list) currently being used?

I don't know. It is in pyicloud code that was grandfathered

> 
> Now the exact same thing applies to `extendedDescEnc` and `keywordsEnc`.

if there is existing code that you do not know the use of, it is not a reason on itself to add more code that is not used. Provide code to `icloudpd` that will use these new fields and improve the app in a meaningful way

AndreyNikiforov (changes requested):
Provide code to icloudpd that will use these new fields and improve the app in a way meaningful to the users.

PR #924, opened 2024-07-30
fix problem : photos of Recently Deleted won't be expunged with   --album "Recently Deleted"  --delete-after-download

xbd-gh:
when expunge photos of Recently Deleted , the post date should with "fields": {"isExpunged": {"value": 1}} not "fields": {"isDeleted": {"value": 1}}

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -1018,7 +1018,7 @@ def delete_photo(
     post_data = json.dumps(
         {
             "atomic": True,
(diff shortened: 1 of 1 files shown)


AndreyNikiforov:
See [contributing guidelines](https://github.com/icloud-photos-downloader/icloud_photos_downloader/blob/3cc7958a719aa984b668f8852f59bb1a0b8d1015/CONTRIBUTING.md) for suggestions on running test/lint/etc locally.

PR #931, opened 2024-08-09
 fix problem : photos of Recently Deleted won't be expunged with --album "Recently Deleted“ --delete-after-download

xbd-gh:
when expunge photos of Recently Deleted , the post date should with "fields": {"isExpunged": {"value": 1}} not "fields": {"isDeleted": {"value": 1}}

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -1018,7 +1018,7 @@ def delete_photo(
     post_data = json.dumps(
         {
             "atomic": True,
-            "desiredKeys": ["isDeleted"],
+            #"desiredKeys": ["isDeleted"],#this is not needed,2024-07-29
             "operations": [
                 {
                     "operationType": "update",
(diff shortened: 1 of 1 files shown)


AndreyNikiforov:
Is Recently Deleted size counted towards iCloud storage limit? What is the purpose of downloading Recently Deleted album with `--delete-after-download` flag? IIUC most ppl use that flag to limit iCloud storage (e.g. keep it within free amount).

xbd-gh:
first,the storage of Recently Deleted in icloud is unlimited.free icloud user only have 5GB,when the storage of iphone is near full,they want move some media data to the PC, but the icloud is full,so,they can delete them on the iphone,then downoad from the Recently Deleted album. and delete after download.when the iphone full again,they repeat the operation,if the photo were not deleted then they will be downoaded again.

AndreyNikiforov:
we have [`--keep-icloud-recent-days` parameter](https://icloud-photos-downloader.github.io/icloud_photos_downloader/reference.html#keep-icloud-recent-days-parameter) to control iCloud storage now. See if that solves your problem.

I am not certain yet if expunging deleted photos will not add more confusion and unexpected behavior.

PR #962, opened 2024-10-12
chore: support py3.13 build

chenrui333:
👋 update to support py3.13 build.

relates to https://github.com/Homebrew/homebrew-core/pull/193804


AndreyNikiforov:
to support 3.13, we need to run tests in that version too. Probably makes sense to make 3.13 a default version - change dev image, use it in build images. need to mention in changelog too, so ppl reviewing next release are aware of changes

AndreyNikiforov (changes requested):
Add tests for 3.13 and switch dev env to 3.13

Issue #789, opened 2024-02-15
Download EXIF Favorite flag


iansewell:
## Summary
If I favorite (heart) a photo on my iphone, I'd like to be able to see that metadata in the downloaded photo (via exiftool)

## Context
When I take a bunch of photos instead of using an album for the best photos, I use the built in "heart" mechanism in iOS but these are not downloaded.  It would also be nice if favoriting an already downloaded photo would be "seen" and cause a re-download

AndreyNikiforov:
Curious, once you download an asset with exif favorite flag, how are you going to use it? Does your photo organizing software support it? Do you know which flag that software uses [out of all available](https://exiv2.org/tags.html)?

iansewell:
Hi Andrey, thanks for the response.
I use immich/Immich GO.  TBH I'm not sure what EXIF flag they use - maybe none yet.   My flow is that I use icloudpd to download images which are then uploaded with immich-go to my server.  Yes, I know that Immich has a native IOS client that uploads images directly from the iphone, but it doesn't work which large icloud libraries (even if I seed Immich with a library thanks to icloudpd, dedup'ing from the app takes forever, and I don't think it even works for images that have been offloaded to icloud)

If immich doesn't support the concept of favorites then I could still use a script to separate out the favorited photos using whatever exif flag you use and then upload those to a specific album which could be bulk favorited in immich.

AndreyNikiforov:
If your tool does not support EXIF flag for favorites, then I do not see a reason to implement it in icloudpd yet. iCloud may have a capability to put favorites in separate album automatically, in which case you can use icloudpd to download that album and use immich with that album natively.

philjak:
I'd be also interested in this information. Either via exif or via an extension to the file name. If we'd have this information, I'm pretty sure we could script something to mark those pictures as favorite in Immich.

Issue #448, opened 2022-06-22
Write exif informations from apple metadata


Se7enair:
## Summary
[NOTE]: # ( Provide a brief overview of what the new feature is all about )
Adding metadata from apple photos to exif

## Context
[NOTE]: # ( Why does this feature matter to you? What unique circumstances do you have? )
Apple or the photos app is generating metadata which contains keywords, persons, GPS location, etc. At the moment not all of them are written to the exif information in a picture.
It would be nice, if there is a option like `--exif` to write all these data to the files.
So other programs can make use of it. For example, as far as I know Adobe Lightroom is working with these, or a batch rename/move can use this.

AndreyNikiforov:
What specific metadata attribute(s) from iCloud you would like to see in exif?

Can you please outline a scenario to test functionality once implemented? E.g. 1) upload photo without XYZ exif 2) check that iCloud Web UI shows ABC attribute 3) download photo and check that XYZ matches ABC data.

Se7enair:
In my case I am interested in detected persons and albums.
Persons because it would be easy to search for pictures from a specific person, even on my local files. 
albums to be able to sort the exported pictures in albums. At the moment I only can export a file structure based on the date. 
Or any other thing you can do in combination with exiftool.
At the moment there are exif informations in a exported photo, but as far as I See, albums and persons not.

AndreyNikiforov:
> Persons because it would be easy to search for pictures from a specific person, even on my local files.
albums to be able to sort the exported pictures in albums.

What tool are you using to search local files by exif tags? What tags does that tool use for Person & Album? I scanned https://exiv2.org/tags.html and don't see anything close to the requested metadata.

Se7enair:
At the moment I export from iCloud with [osxphotos](https://github.com/RhetTbull/osxphotos) it writes the information at various tags

![image](https://user-images.githubusercontent.com/1680106/175831934-919073b4-3e6a-46d5-8384-440ef8577180.png)
![image](https://user-images.githubusercontent.com/1680106/175831945-ecee5bf8-b276-4eac-a928-fbea2f98f535.png)
![image](https://user-images.githubusercontent.com/1680106/175831954-c1b132e1-39aa-432f-8ca9-ffb5df41d65b.png)

Christoph Drexler is a person, icloudpd is a album.

[...]

Issue #102, opened 2018-09-23
Add keywords to image metadata


artemryzhov:
Hi! 

I want to implement feature that add keywords to image metadata, by icloud album title. For example, I have 10 photos in album named "Test album", and I want to images have keyword "Test album", for quick search in Shotwell.

The difficulty in implementing data writing to a file..

I tried via **piexif**:

```
  exif_dict = piexif.load(path)
  exif_dict.get("0th")[40094] = tags // tags is string 'tag1,tag2'
  exif_bytes = piexif.dump(exif_dict)
  piexif.insert(exif_bytes, path)       
```

and got error

```  File "/usr/local/lib/python2.7/dist-packages/piexif/_dump.py", line 63, in dump
    zeroth_set = _dict_to_bytes(zeroth_ifd, "0th", 0)
  File "/usr/local/lib/python2.7/dist-packages/piexif/_dump.py", line 337, in _dict_to_bytes
    offset)
  File "/usr/local/lib/python2.7/dist-packages/piexif/_dump.py", line 193, in _value_to_bytes
    value_str = (_pack_byte(*raw_value) +
  File "/usr/local/lib/python2.7/dist-packages/piexif/_dump.py", line 162, in _pack_byte
    return struct.pack("B" * len(args), *args)
struct.error: cannot convert argument to integer
```

via **iptcinfo**:

``` iptc = iptcinfo.IPTCInfo(path)    
        iptc.keywords = []
        tags = [x.strip() for x in tags.split(',')];
        iptc.keywords.extend(tags)
        iptc.save()```

no errors, and no results - tags wasn't written to file...

Help, please...

ndbroadbent:
Hi, wow that sounds difficult! Sorry but I've never really used the piexif
or iptcinfo libraries, so I won't be able to help very much. (Someone else
contributed the piexif code.) But it sounds like a great idea for a
feature, and I'd be happy to merge a PR if you can get it working.

adamshand:
@artemryzhov did you ever find a way of doing this?  I'm loving this as a way to get my photos out of iCloud, but losing 15 years worth of albums is a bummer. :-)

RhetTbull:
[...]

PR #984, opened 2024-11-01
Add Export of metadata via XMP Sidecar

chkuendig:
There have been multiple requests over time ( #448, #102, #789, probably more) to export some of the metadata that can be changed in Photos. This PR tries to address as many of these as possible but focused on the most common ones:
- Time adjustments
- Location Adjustments
- Hidden and Deleted Photos
- Title and Description
- Changed Rotation
- Tags/Keywords

I tested this on my 50'000+ images library and it processed all of them. If the Exif Tags are supported by Immich they also match. (this helped a ton to sort the images in the correct order and add missing location tags to pictures taken without embedded location info - e.g. from DSLR etc)

PS: This also would replace/close #915 I guess.


AndreyNikiforov:
Thanks for taking initiative. Appreciate writing tests for new features.

Can you implement all these features without external dependency on exiftool, pls? `icloudpd` is distributed as standalone binary without any dependencies on all platforms and there is value in keeping it that way.

chkuendig:
I understand the preference to keep this working without any external dependencies. Unfortunately for XMP handling I couldn't find a single library that wouldn't need external dependencies. I specifically looked at:
- [pyexiftool](https://pypi.org/project/PyExifTool/) - needs `exiftool`.
- [python-xmp-toolkit](https://pypi.org/project/python-xmp-toolkit/) - wraps [exempi](https://libopenraw.freedesktop.org/exempi/) but doesn't bundle it. So the library has to be installed separately. It also [doesn't seem to support Windows](https://python-xmp-toolkit.readthedocs.io/en/latest/installation.html).
- [pyexiv2](https://pypi.org/project/pyexiv2/) wraps exiv2 (which it ships with) but [requires multiple shared libraries](https://github.com/LeoHsiao1/pyexiv2/blob/master/docs/Tutorial.md) (`glibc`, `gettext`, `inih`, `Microsoft Visual C++ 2015-2019` depending on platform)

I couldn't find a single pure python library and I don't really have the bandwith to implement a fairly complex standard from scratch just for this. If any of the other 2 libraries are okay, I'm happy to re-implement the PR. Or maybe there's a better way I'm not aware of?

AndreyNikiforov:
is it reasonable too generated xmp file just from metadata received from Apple (without exif at all)? xmp will have to bed generated with the code i assume

chkuendig:
Thats what I'm doing, I use exiftool just to write the sidecar file. 

I'll check if its feasible to just write the XML by hand. As long as we stick to the +/- 10 fields, its cumbersome but probably doable by just generating the xml once with exiftool and then write code to replicate it in pure python.

AndreyNikiforov:
what happens when metadata is updated in icloud? should local xmp be synchronized? what if local changed as well?

chkuendig:
I currently assumed this to be mostly an archival function, so the iCloud data would overwrite local data.

Thinking more about this, I'd suggest we check the [XMLToolkit](https://exiftool.org/TagNames/XMP.html#x) tag to make sure the last edit was by icloudpd. [It seems to be](https://exiftool.org/forum/index.php?topic=15047.0) a tag thats changed if edited by another software.


I'll add this with the update to remove external dependencies.

chkuendig:
I pushed an update that removes the dependency on exitfool and also checks if an xmp file exists before overwriting it (it only overwrites if the file is empty or was generated by icloudpd in the first place)

AndreyNikiforov:
> note: I updated the lint python version to 3.9 so I can use [ElementTree.indent](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.indent)

Does that mean we have to drop support for 3.8?

chkuendig:
Yes. If thats an issue i'd suggest to remove that call and output unformatted xml instead.

AndreyNikiforov:
> Yes. If thats an issue i'd suggest to remove that call and output unformatted xml instead.

Thanks! I think maintaining 3.8 compatibility is higher value than pretty printed small xml file (which is not supposed to be viewed by humans)

chkuendig:
Done.

Issue #1005, opened 2024-11-26
Authentication persistence (bug?)


sdebruyn:
It would be very useful if icloudpd could persist authentication tokens and sessions. I would like to be able to restart the process without having to provide password and/or MFA authentication every time.

I am using Docker and I can persist the password by mounting /root/.local/share/python_keyring/keyring_pass.cfg into a volume. For others reading this, it should be noted that this is not more secure than just putting your Apple ID password in plain text in your Docker config as this contains a base64 encoded string with your password.

But then after persisting my password, I still have to enter my MFA code every time, even though the one I entered previously is probably still valid.

Ideally I would like to persist neither of those 2 (persisting the MFA isn't useful anyway since it's only valid for 30 seconds) but instead just persist a token/session somewhere which can be used to communicate to iCloud without asking me for my credentials.

I thought this is what `--cookie-directory` would do, but that doesn't work. My cookies are properly stored in that folder, but icloudpd is still asking me to re-authenticate every time I start up a container with that folder mounted.

sdebruyn:
After mounting both the keyring dir and the cookies to a volume, it seems to work. Without the mounted keyring, it asks for a password. This is my Docker setup now:

```yaml
version: "3.8"
services:
  icloudpd-sam:
    image: icloudpd/icloudpd:latest
    container_name: icloudpd_sam
    network_mode: bridge
    environment:
      - TZ=Europe/Brussels
    volumes:
      - /volume1/icloudpd/sam:/data
      - cookies:/cookies
      - keyring:/root/.local/share/python_keyring
    command:
      - icloudpd
      - --cookie-directory
      - /cookies
      - --directory
      - /data
      - --username
      - REDACTED
      - --watch-with-interval
      - "600"
      - --set-exif-datetime
      - --delete-after-download
      - --no-progress-bar
      - --password-provider
      - keyring
      - --password-provider
      - webui
      - --mfa-provider
      - webui
    ports:
      - 38080:8080
    restart: on-failure
  icloudpd-sam-notifier:
    image: icloudpd/icloudpd:latest
    container_name: icloudpd_sam_notifier
    network_mode: bridge
    environment:
      - TZ=Europe/Brussels
    volumes:
      - cookies:/cookies
      - keyring:/root/.local/share/python_keyring
    command:
      - icloudpd
      - --cookie-directory
      - /cookies
      - --username
      - REDACTED
      - --auth-only
      - --smtp-username
      - REDACTED
[...]

PR #1006, opened 2024-11-26
pass log_level to pyicloud library

### LINKED ISSUE #123: Log level options
It looks like, though the documentation for icloudpd on GitHub suggest that there's an "error" option for log-level, that it doesn't function and instead outputs debug and info messages as well.

Being a novice coder at best, I'm not quite sure, but when poking around icloud_photos_downloader/icloudpd/logger.py, it looks like there's actually no "error" option.

Is that correct? If so, should/can there be one?

sdebruyn:
Probably resolves #123 

I would also like to see logs from the pyicloud library.

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -621,6 +621,7 @@ def main(
         stream=sys.stdout,
     )
     logger = logging.getLogger("icloudpd")
+    pyicloud_logger = logging.getLogger("pyicloud_ipd")
     if only_print_filenames:
         logger.disabled = True
     else:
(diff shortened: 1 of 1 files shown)


AndreyNikiforov:
Were you able to solve any problem with extra logging yet (or working on something where it helps)? 

What does pyicloud expose with [default DEBUG level](https://github.com/icloud-photos-downloader/icloud_photos_downloader/blob/master/src/icloudpd/base.py#L464)? Does that extra info (if any) improve user experience for everybody using default logging setting?

sdebruyn:
One of the things it tells you is if found a cookie or not, which might be useful. E.g. in the case I had in #1005

AndreyNikiforov:
Does default DEBUG level changes experience for user who do not need extra logs?

sdebruyn:
Probably, but then again, should the default logger level really be DEBUG? Most application would put this on info/warn

AndreyNikiforov (changes requested):
Default log level is DEBUG. Simply turning logging for pyicloudpd_ipd will increase amount of output for all users, 99.99% of which do not need extra logging. This challenge needs to be solved before we can merge changes.

Issue #911, opened 2024-07-13
MFA email notification prevents MFA verification code in WebUI


S0urceror:
## Overview
When smtp_username is specified the MFA dialogue in the WebUI only asks for password but never for verification code. When you delete the smtp_username it does work.

## Steps to Reproduce
1. set smtp_username, password, host and port 
2. set mfa-provider to webui
3. restart docker container

## Expected Behavior
First ask for a password. Then ask for the verification code.

## Actual Behavior
Only asks for a password and never asks for the verification.

## Context
The code contains:
```
        if icloud.requires_2fa:
            **if raise_error_on_2sa:
                raise TwoStepAuthRequiredError("Two-factor authentication is required")**
            logger.info("Two-factor authentication is required (2fa)")
            if mfa_provider == MFAProvider.WEBUI:
                request_2fa_web(icloud, logger, status_exchange)
            else:
                request_2fa(icloud, logger)

        elif icloud.requires_2sa:
            if raise_error_on_2sa:
                raise TwoStepAuthRequiredError("Two-step authentication is required")
            logger.info("Two-step authentication is required (2sa)")
            request_2sa(icloud, logger)
```
I believe that you should not raise the 2SA error on 2FA request. And raising an exception prevents from running request_2fa_web.

holomekc:
See #898

AndreyNikiforov:
1.29.1 should send notifications without terminating the program

PR #1007, opened 2024-11-26
fix SMTP notification not sent

### LINKED ISSUE #898: service exits after sending 2fa notification email on startup

sdebruyn:
Fixes #898 #911

Before attacking the tests, let me know what you think of this approach.

Raising the exception will always exit the current code path in authentication.py

### DIFF
--- a/src/icloudpd/authentication.py
+++ b/src/icloudpd/authentication.py
@@ -14,13 +14,6 @@ from icloudpd.mfa_provider import MFAProvider
 from icloudpd.status import Status, StatusExchange
 
 
-class TwoStepAuthRequiredError(Exception):
-    """
-    Raised when 2SA is required. base.py catches this exception
-    and sends an email notification.
-    """
-
-
 def authenticator(
     logger: logging.Logger,
     domain: str,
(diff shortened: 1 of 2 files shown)


sdebruyn:
As long as the authenticator is called, it will also trigger the callback again. And indeed, it keeps it from exiting after sending the email. I also needed this to work better for #1008

Issue #975, opened 2024-10-25
401 response when authenticating with version 1.24.0


scttnlsn:
## Overview

401 response when authenticating with version 0.1.24

## Steps to Reproduce

1. Install version `0.1.24`
2. Run `icloudpd --username '<email>' --password '<password>' --auth-only`

## Expected Behavior

Authentication succeeds.

## Actual Behavior

Authentication fails with this error message:

```
pyicloud_ipd.exceptions.PyiCloudAPIResponseException: {
  "serviceErrors" : [ {
    "code" : "-20101",
    "message" : "Enter the email or phone number and password for your Apple Account.",
    "suppressDismissal" : false
  } ]
} (401)
```

## Context

I was experiencing the authentication issues described in https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/970.  Tried upgrading to 0.1.24 (which includes a fix for that) but I'm getting another authentication error.

Jan108:
I have the same problem, but I get a different error code from the API:
```
jan@jan:~/$ ./icloudpd-ex-1.24.0-linux-amd64 icloudpd --username '<iCloud email>' --auth-only
2024-10-26 13:46:24 DEBUG    Authenticating...
iCloud Password: 
Traceback (most recent call last):
  File "pyicloud_ipd/base.py", line 270, in authenticate
pyicloud_ipd.exceptions.PyiCloudAPIResponseException: {
  "serviceErrors" : [ {
    "code" : "-20283",
    "message" : "Enter the email or phone number and password for your Apple Account.",
    "suppressDismissal" : false
  } ]
} (401)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "starters/icloudpd_ex.py", line 122, in <module>
  File "starters/icloudpd_ex.py", line 118, in main
  File "click/core.py", line 1157, in __call__
  File "click/core.py", line 1078, in main
  File "click/core.py", line 1688, in invoke
  File "click/core.py", line 1434, in invoke
  File "click/core.py", line 783, in invoke
  File "icloudpd/base.py", line 745, in main
  File "icloudpd/base.py", line 1183, in core
  File "icloudpd/authentication.py", line 53, in authenticate_
  File "pyicloud_ipd/base.py", line 160, in __init__
  File "pyicloud_ipd/base.py", line 273, in authenticate
[...]

PR #1018, opened 2024-12-09
fix: fallback to old raw password auth if srp auth fails

iowk:
As mentioned by @gcobb321 https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/975#issuecomment-2510446817, Apple seems to be rolling back to the old non-SRP authentication for some accounts.

This PR creates a flow to fallback to the old authentication when SRP authentication fails.

I have an account that failed to login with SRP, and is fixed with this PR.

### DIFF
--- a/src/pyicloud_ipd/base.py
+++ b/src/pyicloud_ipd/base.py
@@ -192,93 +192,15 @@ class PyiCloudService:
 
         if not login_successful:
             LOGGER.debug("Authenticating as %s", self.user["accountName"])
-
-            headers = self._get_auth_headers()
-
-            class SrpPassword():
-                # srp uses the encoded password at process_challenge(), thus set_encrypt_info() should be called before that
-                def __init__(self, password: str):
-                    self.pwd = password
-
-                def set_encrypt_info(self, protocol: str, salt: bytes, iterations: int) -> None:
-                    self.protocol = protocol
-                    self.salt = salt
-                    self.iterations = iterations
-
-                def encode(self) -> bytes:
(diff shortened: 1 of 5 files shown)


src/pyicloud_ipd/base.py
```diff
@@ -192,93 +192,15 @@ def authenticate(self, force_refresh:bool=False, service:Optional[Any]=None) ->
...
-                        data=json.dumps({}),
-                        headers=headers,
-                    )
-                elif response.status_code >= 400 and response.status_code < 600:
-                    raise PyiCloudAPIResponseException(response.text, str(response.status_code))
-            except PyiCloudAPIResponseException as error:
-                msg = "Invalid email/password combination."
-                raise PyiCloudFailedLoginException(msg, error) from error
+                self._authenticate_srp()
+            except PyiCloudFailedLoginException as error:
```
AndreyNikiforov, line 197: I wonder if there is anything in payload we can use as a flag for srp rather than try-catch.

If try-catch is the only way, what other exceptions may be incorrectly channelled to non-srp path?

iowk: For the account I tried, the response for srp signin/complete is:
```
('Invalid email/password combination.', PyiCloudAPIResponseException('{\n  "serviceErrors" : [ {\n    "code" : "-20283",\n    "message" : "Enter the email or phone number and password for your Apple\xa0Account.",\n    "suppressDismissal" : false\n  } ]\n} (401)'))
```

Which is the same as when the password is incorrect.

This also means if the password is incorrect, it will be incorrectly channelled to non-srp path.

PR #1022, opened 2024-12-18
Added option to download X oldest items, not only X recent ones (building upon Santa77's #620)

lcorbasson:
Added option to download X oldest items, not only X recent ones (see PR #620 by Santa77).

Synced with the last version in master.


AndreyNikiforov:
What is the use case for this functionality? I can see how "--recent" helps with optimizing incremental updates (and with testing), but I fail to see usage of the "--oldest" flag.

lcorbasson:
This is useful to delete the oldest pictures in the iCloud library while keeping the most recent ones.

roju:
My iCloud storage is full because there are too many old photos. I want to download some of these old photos, and then delete after download. I also want to keep a certain number of recent photos in my iCloud account at all times.

AndreyNikiforov:
> This is useful to delete the oldest pictures in the iCloud library while keeping the most recent ones, e.g.: `icloudpd -d "$ICLOUD_PICTURES_LOCAL_DIRECTORY" -u "$ICLOUD_USERNAME" --delete-after-download --oldest 20000`

That is quite useful! I haven't though about such use case.

src/icloudpd/base.py
```diff
@@ -1289,6 +1299,13 @@ def core(
                 photos_count = recent
                 photos_enumerator = itertools.islice(photos_enumerator, recent)
 
+            # Optional: Only download the x oldest photos.
+            if oldest is not None:
+                photos_count = oldest
+                total_photos = len(photos_enumerator)
+                start = total_photos - oldest
+                photos_enumerator = itertools.islice(photos_enumerator, start, total_photos)
```
AndreyNikiforov, line 1307: I am curious how that works on large collections in iCloud.com. I _think_ we are processing images as a stream and any filtering is done on a fly instead of memory.

AndreyNikiforov (changes requested):

AndreyNikiforov:
> This is useful to delete the oldest pictures in the iCloud library while keeping the most recent ones, e.g.: `icloudpd -d "$ICLOUD_PICTURES_LOCAL_DIRECTORY" -u "$ICLOUD_USERNAME" --delete-after-download --oldest 20000`

@lcorbasson There may be another option to solve your use case: tweak delete-after-download option. E.g. "Delete after downloading if older than 30days from now except favorites". That may be useful to broader audience who keep iCloud.com storage to min. It may worth discussing all edge cases and analyze reported issues before coding though.

jhofker:
Check our #1040 that was just merged and the larger conversation in #1046 that may be of-interest here!

Issue #1032, opened 2024-12-28
After first run on Ubuntu fails with "keyring.errors.KeyringLocked: Failed to unlock the collection"


Gorbush:
## Overview
```shell
$ uname -a
Linux 5.15.0-127-generic #137-Ubuntu SMP Fri Nov 8 15:21:01 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
$  icloudpd --version
version:1.25.0, commit sha:96f9451, commit timestamp:Tue Dec  3 12:01:43 2024 EST
```

## Steps to Reproduce
[NOTE]: # ( Provide a simple set of steps to reproduce this bug. Version of the app, OS/System, and minimum command line to reproduce. )
1.  Run the download command as recommeneded with my user name 
```shell
icloudpd --directory /data/to_import/iCloud-User --username user@gmail.com --watch-with-interval 3600
```
2. First run was correct, it started downloading the photos.
3. I had to turn the PC off, and next morning I tried to re-start the command to continue downloading, but it fails with exception.

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen )
Re-authenticates and continue the download of phtos.

## Actual Behavior
[NOTE]: # ( Tell us what actually happens )
Fails with exception:
```shell
2024-12-28 10:33:50 DEBUG    Authenticating...
Traceback (most recent call last):
  File "starters/icloudpd.py", line 6, in <module>
  File "click/core.py", line 1157, in __call__
  File "click/core.py", line 1078, in main
  File "click/core.py", line 1434, in invoke
  File "click/core.py", line 783, in invoke
  File "icloudpd/base.py", line 753, in main
  File "icloudpd/base.py", line 1196, in core
  File "icloudpd/authentication.py", line 51, in authenticate_
  File "pyicloud_ipd/utils.py", line 37, in get_password_from_keyring
  File "keyring/core.py", line 56, in get_password
  File "keyring/backends/chainer.py", line 49, in get_password
  File "keyring/backends/SecretService.py", line 79, in get_password
  File "keyring/backends/SecretService.py", line 68, in get_preferred_collection
keyring.errors.KeyringLocked: Failed to unlock the collection!
[84394] Failed to execute script 'icloudpd' due to unhandled exception!
```

## Context
[NOTE]: # ( Tell us why expected behvior is valuable )

[NOTE]: # ( Give us any additional information you may have. )

zhaobenx:
same on debian 12 

[...]

PR #1033, opened 2024-12-28
Fix for issue #1032 - when keyring is for some reason broken or fail - skips it to proceed with next method of authentication

Gorbush:
Helps with https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/1032


Gorbush:
Working on test

PR #1034, opened 2024-12-29
Add curl to Dockerfile

kusold:
When mounting a notification script in the docker container, curl is unavailable which limits most use cases. Busybox's wget is preinstalled, but it doesn't support authentication.

This PR add curl to all the images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,18 @@
 FROM alpine:3.18 AS runtime_amd64_none
 ENV MUSL_LOCPATH="/usr/share/i18n/locales/musl"
-RUN apk update && apk add --no-cache tzdata musl-locales musl-locales-lang
+RUN apk update && apk add --no-cache tzdata musl-locales musl-locales-lang curl
 WORKDIR /app
 COPY dist/icloudpd-ex-*.*.*-linux-musl-amd64 icloudpd_ex
 
 FROM alpine:3.18 AS runtime_arm64_none
 ENV MUSL_LOCPATH="/usr/share/i18n/locales/musl"
-RUN apk update && apk add --no-cache tzdata musl-locales musl-locales-lang
+RUN apk update && apk add --no-cache tzdata musl-locales musl-locales-lang curl
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


AndreyNikiforov:
`icloudpd` docker containers come with no promises for the content of the container other than binaries. tz and locales are accidental promises. I would like to keep promises to minimum to easy future modifications.

Are there any other ways to a) support authentication for http requests from notification scripts OR b) add curl/ssl/etc at use- instead of build- time? I noticed that Devcontainers for vscode have "feature" add-ons, which seems to be containers as well - that may be an example of "merging" two containers.

kusold:
> I would like to keep promises to minimum to easy future modifications

Fair enough. I also prefer slim containers, so I understand your POV.

> Are there any other ways to a) support authentication for http requests from notification scripts

I'm passing `--user 5555:5555` which doesn't have access to install apps via `apk`

> OR b) add curl/ssl/etc at use- instead of build- time

Building a custom container is the route I'll need to go. Luckily `icloudpd` is pretty simple to install with the minimal dependency set.

----

Thank you for creating this application. It's becoming an important part of my backup strategy.

AndreyNikiforov:
> `icloudpd` docker containers come with no promises for the content of the container other than binaries

for posterity: Actual [promise](https://icloud-photos-downloader.github.io/icloud_photos_downloader/install.html#docker) is that two commands are supported, not the presence of the particular binaries. I am sorry I mislead you.

Docker was added to the project much later than calls to notification scripts, so these two approaches are not perfectly aligned.

boredazfcuk:
@kusold You don't need to build a custom container... just download the latest curl binary (https://github.com/moparisthebest/static-curl/releases/download/v8.11.0/curl-amd64) to somewhere like /opt/curl-amd64. Make it executable with `chmod +x /opt/curl-amd64` then map it into your container with `-v /opt/curl-amd64:/bin/curl` e.g.:
```
docker run -it --rm --name icloudpd_test2 -v /opt/curl-amd64:/bin/curl -v $(pwd):/data -e TZ=America/Los_Angeles icloudpd/icloudpd:latest icloudpd --directory /data --username my@email.address --watch-with-interval 3600
```

Issue #174, opened 2020-10-16
Keep photos from last 30 days


konradorlinski:
Hi!
Is there any posibility to download all photos with delete option, but don't delete photos from 30 days or keep e.g. last 100 photos?

menkej:
This is not implemented yet. I'm also a little concerned about deleting the users photos. Imagine you are just "playing around with options" and delete your pictures by accident... I would leave this feature out and let the user delete his pics himself if he/she wants to...

konradorlinski:
Thanks for answer!
I propably wrong understood autodelete option. Now all is clear. Indeed, delete feature will be propably unsafe

PR #1040, opened 2025-01-01
feat: Add flag --keep-icloud-recent-days

jhofker:
In #659 and #174 (and kind of #820) there's a desire to copy all photos and _move_ all photos older than a certain timeframe. This would allow users to keep some photos in their iCloud for easy sharing/viewing, but also maintain their storage in a way that works for them.

This new flag only works when `--delete-after-download` is also set, otherwise it does nothing.

I could use a hand generating a cassette file that actually allows for testing this as I'm trying not to use personal photos as recommended in the docs. 😃


jhofker:
@AndreyNikiforov do you have any advice on properly generating a cassette to test these scenarios? I assume the account specified in each existing test is a fake, but I could be wrong.

AndreyNikiforov:
> @AndreyNikiforov do you have any advice on properly generating a cassette to test these scenarios? I assume the account specified in each existing test is a fake, but I could be wrong.

see [doc](https://github.com/icloud-photos-downloader/icloud_photos_downloader/blob/master/CONTRIBUTING.md#how-to-write-a-unit-test) for some basics. Generating vcr is a pain by my books. For small cases I cloned existing files and changed them manually. 

I hope to separate IO from filtering and decision logic, so we can keep vcr for IO portion only. We are not there yet.

jhofker:
Glad to know it's not just me 😃 

I updated the test to reuse the cassette from the base delete_after_download and mocked the current datetime so that we can ensure the recent days check is working.

I do see two other tests failing, but they also fail on `master`:
```
FAILED tests/test_autodelete_photos.py::AutodeletePhotosTestCase::test_autodelete_photos_lp - AssertionError: 'INFO     Deleted tests/fixtures/test_autodelete_photos_lp/data/2018/07/12/IMG_7190.JPG' not found in "DE...
FAILED tests/test_autodelete_photos.py::AutodeletePhotosTestCase::test_autodelete_photos_lp_heic - AssertionError: 'INFO     Deleted tests/fixtures/test_autodelete_photos_lp_heic/data/2018/07/12/IMG_7190.HEIC' not found ...
```

@AndreyNikiforov if there are any other changes you'd like to see in this, just let me know!

AndreyNikiforov:
@jhofker pls see if #1046 helps

jhofker:
👀 I'll check it out and add comments over there!

jhofker:
Yep -- big fan.

Happy to do any of the following:
- rename the parameter in this PR to `keep-cloud-recent-days` to push forward
- turn this PR over to you
- close this PR and be patient (or help)

Let me know what works best for you.

src/icloudpd/base.py
```diff
@@ -488,7 +488,7 @@ def report_version(ctx: click.Context, _param: click.Parameter, value: bool) ->
     is_flag=True,
 )
 @click.option(
-    "--keep-recent-days",
+    "--keep-icloud-recent-days",
     help="When using --delete-after-download, do not delete photos newer than this many days",
```
AndreyNikiforov, line 492: We are using `-keep-icloud-recent-days` _instead_ of `--delete-after-download` according to RFC #1046. Also need to check that both params are never used together.

src/icloudpd/base.py
```diff
@@ -488,7 +488,7 @@ def report_version(ctx: click.Context, _param: click.Parameter, value: bool) ->
...
     type=click.IntRange(0),
     default=0,
```
AndreyNikiforov, line 494: We do not need default equivalent of wiping the whole iCloud collection. Without default the value of the param is None.

src/icloudpd/base.py
```diff
@@ -1364,17 +1364,17 @@ def should_break(counter: Counter) -> bool:
                         break
                     item = next(photos_iterator)
                     if download_photo(consecutive_files_found, item) and delete_after_download:
```
AndreyNikiforov, line 1366: IIUC download_photo returns True only if file was actually downloaded. We need to support deleting even if file already exists.

Beter to have test for deleting existing file.

tests/test_download_photos.py
```diff
@@ -1453,7 +1452,7 @@ def test_download_after_delete_fail(self) -> None:
         # TODO assert cass.all_played
         assert result.exit_code == 0
 
-    def test_delete_after_download_keep_recent_days(self) -> None:
+    def test_delete_after_download_keep_icloud_recent_days(self) -> None:
```
AndreyNikiforov, line 1455: Need another test for existing file.

tests/test_download_photos.py
```diff
@@ -1485,7 +1484,7 @@ def test_delete_after_download_keep_recent_days(self) -> None:
                             "--threads-num",
                             "1",
                             "--delete-after-download",
-                            "--keep-recent-days",
+                            "--keep-icloud-recent-days",
                             "5",
```
AndreyNikiforov, line 1488: It may be more valuable to check edge of the interval instead of definate overlapping: Running on 8/2/2018 with `--keep-icloud-recent-days 2` should keep img from 7/31/2018, but running on 8/2/2018 with `--keep-icloud-recent-days 1` should not keep img from 7/31/2018, right?

Running on 7/31/2018 with `--keep-icloud-recent-days 0` should not keep image from 7/31/2018 as its age < 1 days, right?

AndreyNikiforov (changes requested):
Pls add a line to changelog.md as well.

src/icloudpd/base.py
jhofker: Updated help. Added check below for the flags being used together.

src/icloudpd/base.py
jhofker: Updated default. Initial check below was already checking for `None`. Good call!

tests/test_download_photos.py
jhofker: Moved to separate file

tests/test_download_photos.py
AndreyNikiforov: > Yeah, those sound right to me. However, since the calculation is done with whole days, depending on the time it was created on the day, it may or may not get picked up by the check. Personally I'm okay with that, since I'm running this twice a day and in the long run, I just want to keep "roughly" the number of days I'm specifying. How do you feel about the _date_ matching but the photo not being deleted because it's 25 hours old in some instances?

I like to keep days == 24hrs and use full timestamps for calcs. Using just dates requires proper TZ handling and can get tricky.

We can technically allow suffix for days and hrs, but that is extra complexity for just a narrow edge case that barely anybody faces.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.25.2 (2025-01-07)
```
AndreyNikiforov, line 5: remove release line. all new features are added to "unreleased" section. I add release and date when I publish - maintaining semantic version based on the set of changes.

tests/test_keep_icloud_mode.py
```diff
@@ -0,0 +1,226 @@
...
+                    )
+                    self.assertIn("INFO     All photos have been downloaded", self._caplog.text)
+                    assert result.exit_code == 0
+
+    def test_narrow_range_keep_icloud_recent_days(self) -> None:
+        base_dir = os.path.join(self.fixtures_path, inspect.stack()[0][3])
+
+        files_to_download = [("2018/07/31", "IMG_7409.JPG")]
+
+        with mock.patch.object(piexif, "insert") as piexif_patched:
```
AndreyNikiforov, line 80: why do we need exif? we use metadata returned in json, so exif is irrelevant.

jhofker: Good call, we don't. This was leftover from copy-pasting a test in the old file to make a new one.

src/icloudpd/base.py
```diff
@@ -1391,6 +1408,40 @@ def should_break(counter: Counter) -> bool:
                     logger, dry_run, library_object, folder_structure, directory, primary_sizes
                 )
 
+            if keep_icloud_recent_days is not None:
+                try:
+                    now = datetime.datetime.now(get_localzone())
+                    created_date = item.created.astimezone(get_localzone())
+                    age_days = (now - created_date).days
+                    if keep_icloud_recent_days != 0 and age_days <= keep_icloud_recent_days:
```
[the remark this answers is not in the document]
AndreyNikiforov: I got days == 0 for close timestamps: 
```
$ python
Python 3.13.1 (main, Dec  4 2024, 20:55:07) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> dt1 = datetime.datetime.now()
>>> dt2 = datetime.datetime.now()
>>> dt1
datetime.datetime(2025, 1, 8, 4, 0, 57, 458939)
>>> dt2
datetime.datetime(2025, 1, 8, 4, 1, 3, 281870)
>>> (dt2-dt1).days
0
>>> 
```
What is different for your trials?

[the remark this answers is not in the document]
jhofker: Ah, only because of photos taken "in the future" -- if I log the times and rerun `test_keep_icloud_recent_days_1_keeps_today` without setting the hour/minute, we get an awkward comparison and resulting log:
```
DEBUG    now: 2018-07-31 00:00:00+00:00, created_date: 2018-07-31 02:22:24.816000-05:00
DEBUG    Skipping deletion of IMG_7409.JPG as it is within the keep_icloud_recent_days period (-1 days old)
```

Which makes sense. And also shouldn't occur in real-world usage.

mzeecedric:
I am wondering if this will work together with the `--auto-delete` flag? What will happen if both is active, e.g.

```
--auto-delete
--keep-icloud-recent-days 1095
```

According to this documentation https://icloud-photos-downloader.github.io/icloud_photos_downloader/mode.html it seems, that only one 'mode' can / should be used.

My intention is to remove old photos from icloud and keep the newer ones in sync (i.e. delete stuff that goes to trash).

jhofker:
`--auto-delete` really just empties the iCloud trash, whereas `--keep-icloud-recent-days` actively deletes items from iCloud for you. 

> My intention is to remove old photos from icloud and keep the newer ones in sync (i.e. delete stuff that goes to trash).

It sounds like `--auto-delete` is the best mode for you, given that!

AndreyNikiforov:
> My intention is to remove old photos from icloud and keep the newer ones in sync (i.e. delete stuff that goes to trash).
> 

I never though of such use case. Kinda maintain synchronized local copy of iCloud for the last X days, but save space in iCloud by deleting older staff. I don't think it is possible to do that today. It may or may not be achievable - need to ensure logic is valid in all edge cases, e.g. never deletes locals older than X days.

> (oh, and what a pity that `--keep-icloud-recent-days` won't work if not rescanning the whole library so I can not use `--until-found` then either.)

I am not following this use case.


It will be better to get visibility for the identified use cases as separate GH issues rather than comments in the closed PR.

Issue #1046, opened 2025-01-06
[RFC] Maintaining iCloud Storage Size


AndreyNikiforov:
# [RFC] Maintaining iCloud Storage Size

This is a proposal to control the size of the iCloud storage by automatically deleting old assets already present on the local storage.

## 1. Context

Many iCloud users store their entire asset collection in the cloud and use the `icloudpd` tool to maintain a local copy. This approach can become less optimal with larger collections due to:

- Increased cost, especially beyond free 5GB tier
- Limited support for assets originating outside of the Apple ecosystem

It is also possible to use the iCloud as an temporary storage between iDevice and a local storage - `icloudpd` has `--delete-after-download` parameter to control deletion of the iCloud assets as soon as they are downloaded to the local storage.

The above are two extremes - "all assets premanently in the cloud" and "no assets permanently in the cloud". There has been a number of requests to support hybrid approach to control the cost of the iCloud service - to keep some assets in the iCloud, while deleteing others. This is the focus of the current proposal.

## 2. Problem Scope

To control the content of the iCloud storage, two priorities have been identified:

1. Limit the iCloud storage by the total size of all assets, e.g. keep latest 5GB of assets to stay within free tier #TBD
1. Limit by the age of the assets, e.g. keep last 30 days of assets to prioritize usefulness #TBD

There is a need for exceptions, such as keeping certain assets (e.g., favorites) in the iCloud permanently. These exceptions will contribute to the total size calculation if applicable. #TBD

## 3. Assumptions

Local storage represents the final storage media for assets. Subsequent runs of the `icloudpd` agains local storage will check for asset presence and only download incremental changes.

Timestamp when the assets was taken (aka "created date") as reported by the iCloud is used for calculating age. For example, asset taken in 2000, but added to the iCloud in 2024, will be 25 years old in 2025. Same timestamp is used for folder structure in the current system.

Assets may be represented in multiple variations, e.g. raw+jpeg, small+medium+original, image+video. Only some variations are counted towards the iCloud total size. Content of smart folders, including Recently Deleted, are not counted towards the total size. This is standard behavior of the iCloud.

### Asset Filters

Asset filters, e.g. `--skip-videos`, limit assets that are checked by the `icloudpd` and only these checked assets are subject for deletion as part of storage maintenence. For example, running `icloudpd` with `--skip-videos` on a huge iCloud collection of videos will download and delete nothing.

[...]

PR #1051, opened 2025-01-10
DRAFT: feat: Add --keep-icloud-album param

jhofker:
Following #1040, this adds the next part of #1046 which allows users to protect specific albums.

Note: @AndreyNikiforov -- when I was merging my first attempt with the changes from #1040, I noticed that the `keep-icloud-recent-days` block was not inside the main iterator loop, which means (😰) that check only operated on the very last item. This should also fix that issue.

This PR currently fails in the `test_keep_icloud_album_exists` test. I _believe_ this is because passing the name of a valid album causes a network request to get details of that album. I'm not sure how to fill out the cassette to actually expect that.

This PR could also use additional tests. I'm currently cheating by selecting `'All Photos'` as the album, and I'd much rather have a good way of verifying the `'Favorites'` album instead.


AndreyNikiforov:
Can you put a separate PR to address the bug pls? With tests. In case we will be discussing new feature for some time, we have bug covered. TIA

src/icloudpd/base.py
```diff
@@ -1360,6 +1373,25 @@ def should_break(counter: Counter) -> bool:
...
+                keep_icloud_album_obj = (
+                    library_object.albums[keep_icloud_album] if keep_icloud_album else None
+                )
+                keep_icloud_album_photo_ids = (
+                    {photo.id for photo in keep_icloud_album_obj.photos}
+                    if keep_icloud_album_obj
+                    else set()
+                )
+            except KeyError:
+                if keep_icloud_album:
```
AndreyNikiforov, line 1387: If user put wrong album name we warn about it and still delete all photos from it. Better user experience would be to check for album and stop action. Probably better to check before we start iteration on photos, so in case `--watch-with-interval` is used, we check for album every interval (e.g. if album was removed in iCloud).

jhofker: Updated this to return early

jhofker:
#1053 is the reduced-scope PR

AndreyNikiforov:
> This PR currently fails in the `test_keep_icloud_album_exists` test. I _believe_ this is because passing the name of a valid album causes a network request to get details of that album. I'm not sure how to fill out the cassette to actually expect that.

You already have a request for All Photos album in your cassette, so the request for any other album will be the same, right?

PR #1053, opened 2025-01-11
bug: keep_icloud_recent days should be inside the photo loop

jhofker:
In #1040 I mistakenly kept the handling of `keep_icloud_recent_days` outside of the main photo iteration, which meant we only "cleaned up" the last photo in the loop.

This moves it inside the loop, then updates a test to verify multiple files will be deleted as expected.

### DIFF
--- a/tests/test_keep_icloud_mode.py
+++ b/tests/test_keep_icloud_mode.py
@@ -5,8 +5,10 @@ from unittest import TestCase, mock
 
 import pytest
 from vcr import VCR
+from requests import Response
 
 from tests.helpers import path_from_project_root, run_icloudpd_test
+from pyicloud_ipd.services.photos import PhotoAsset
 
 vcr = VCR(decode_compressed_response=True, record_mode="none")
 
@@ -153,7 +155,7 @@ class KeepICloudModeTestCases(TestCase):
                 self._caplog.text,
             )
             self.assertIn(
-                f"DEBUG    Deleted IMG_7409.JPG as it is older than the keep_icloud_recent_days period ({days_old - 1} days old)",
+                "INFO     Deleted IMG_7409.JPG in iCloud",
                 self._caplog.text,
             )
             self.assertIn("INFO     All photos have been downloaded", self._caplog.text)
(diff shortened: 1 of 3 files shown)


tests/test_keep_icloud_mode.py
```diff
@@ -208,46 +210,68 @@ def test_keep_icloud_recent_days_1_keeps_today(self) -> None:
...
-                f"INFO     Downloading the first original photo to {data_dir} ...",
-                self._caplog.text,
-            )
-            self.assertIn(
-                f"DEBUG    Deleted IMG_7409.JPG as it is older than the keep_icloud_recent_days period ({days_old - 1} days old)",
-                self._caplog.text,
-            )
-            self.assertIn("INFO     All photos have been downloaded", self._caplog.text)
-            assert result.exit_code == 0
+        def mocked_download(pa: PhotoAsset, _url: str) -> Response:
```
AndreyNikiforov, line 221: Do we really need this since we are creating files before the test?

jhofker: It was the only way I found to get the test working, but again, I'm really bad at these cassette tests 😃 

What I saw when running the tests was that it was attempting to make a network request to get the photos, and those requests weren't matching with requests in the cassette file.

AndreyNikiforov: A cassette is just recorded requests and responses over the wire. If we created 3 files locally and expect delete to be called for some of them, then there should be no download requests made (and cassette should not have download requests), rigth? 

It is possible that cassette has request for list of photos in iCloud that responds with >3 photos and that is why it tries to download "the rest". For this case in other tests I used "--recent" param.

jhofker: Yes, I think so? It's the same data as `list_photos.yml` with an added delete request on the end of it.

I'm using `--recent 3` in this, but if I'm understanding you correctly, that should be `--recent 1` because we don't actually care for it to try to download more than one.

AndreyNikiforov: `--recent 3` filters all icloud content to only 3 records (as if the whole iCloud is just 3 records, even if "list" request in cassette returns 100s). Then for each of these, code checks for existing on local files. We have 3 created as part of the test, so nothing new should be queued for downloading and no downloading requests should be happening. Then all three should be compared for age and some of them may need to be deleted from iCloud. 

Since you are fixing a bug with deletion checked for more than just latest record, you need to filter for more than 1 record, so you cannot do `--recent 1`. You also need to confirm that deletion is running for more that 1 last record, so you need to pick mock date accordingly. For example use 3 records total, mock date & `--keep-icloud-recent-days 1` so that 2 oldest are deleted (because we need to confirm that deletion works for more than 1 record).

WDYT?

jhofker: I swear it didn't work without the mock download before, but removing it now and it passes properly. 

I left this test still 10 days in the future from the files, showing that it will delete everything when the situation calls for it.

I also added another test where keep-icloud-recent-days is 1, which causes the test to delete two files but keep a third. This shows we can iterate over multiple and make different decisions properly based on the days, like you suggested.

How do you feel about these tests?

PR #1095, opened 2025-03-18
feature: the icloud email username is now included in the email about 2sa authentication failing

bed42:
I run icloudpd for multiple iCloud accounts and it's nicer to know which account needs 2fa redone in the email notification itself, rather than having to try and figure it out. This change just adds the email user in an obfuscated manner ie `a****w@icloud.com`:

```
Hello,

a****w@icloud.com's two-step authentication has expired for the icloud_photos_downloader script.
Please log in to your server and run the script manually to update two-step authentication.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@
 
 ## 1.23.0 (2024-07-25)
 
-- feature: update webui and allow to cancel and resume sync
+- feature: update webui and allow to cancel and resume sync 
 - deprecate linux 386 and arm v6 support
 - add linux musl builds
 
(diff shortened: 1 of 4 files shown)


AndreyNikiforov:
Thanks for the change. I understand the use case of maintaining icloudpd for multiple appleids and receiving notification on one email address only (and, subsequently, the need to disambiguate account). However, I do not understand the need to obfuscate appleid - notification is sent to email that already has access to all those appleid, so hiding has no value. Am I missing anything?

bed42:
Just over pre-cautious, I guess as I consider email to be Personally Identifiable Information and email itself to be an insecure medium. Happy to remove if you think it’s too much :)On 19 Mar 2025, at 3:30 am, Andrey Nikiforov ***@***.***> wrote:﻿
  Thanks for the change. I understand the use case of maintaining icloudpd for multiple appleids and receiving notification on one email address only (and, subsequently, the need to disambiguate account). However, I do not understand the need to obfuscate appleid - notification is sent to email that already has access to all those appleid, so hiding has no value. Am I missing anything?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



AndreyNikiforov left a comment (icloud-photos-downloader/icloud_photos_downloader#1095)
Thanks for the change. I understand the use case of maintaining icloudpd for multiple appleids and receiving notification on one email address only (and, subsequently, the need to disambiguate account). However, I do not understand the need to obfuscate appleid - notification is sent to email that already has access to all those appleid, so hiding has no value. Am I missing anything?

AndreyNikiforov:
> Just over pre-cautious I guess, as I consider email addresses to be Personally Identifiable Information and email itself to be an insecure medium. Happy to remove if you think it’s too much :)

We are using email addresses (PII) to identify accounts already disclosed to the recipient of the notification.  Obfuscating will make this use case only partially solved and will adds to maintenance, while value of added security is minuscule if any imo.

I checked that Azure notification are including email addresses used for account identification, so there is a precedent.

Let remove obfuscation until we find actual use case for it.

bed42:
fair evaluation, good thinking checking somewhere like Azure 🤘 Updated.

Issue #787, opened 2024-02-04
Only small number of photos is downloaded from library with 44k photos


skumka:
## Overview
[TIP]:  # ( DO NOT include your login-information when pasting in program output! )
[NOTE]: # ( Give a BRIEF summary about your problem )

## Steps to Reproduce
[NOTE]: # ( Provide a simple set of steps to reproduce this bug. )
1. 
2. 
3. 

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen )

## Actual Behavior
[NOTE]: # ( Tell us what actually happens )

## Context
[NOTE]: # ( Give us any additional information you may have. )

AndreyNikiforov:
cmd `root@docker:/tmp/temp# ./icloudp --username sxxx@xx.xxx  --password :xxx --watch-with-interval 3600 --directory /data` is downloading into `data` folder in the root of your drive, while `ls` cmd later checks `/tmp/temp/data`.

skumka:
Bad copy - it is ./data by mistake removed it when anonymizing pass and user

to be completely sure foleder /data does not exist.
![_2024-02-04_19-00-41](https://github.com/icloud-photos-downloader/icloud_photos_downloader/assets/71942616/fa43768c-7634-431d-8719-c0d352ebd793)

AndreyNikiforov:
Where did you get `./icloudp` binary?

Do you mind trying `docker run --rm -it icloudpd/icloudpd:1.17.3 icloudpd -u your@email.address --recent 2 --dry-run -d ./` pls? Expecting log with two recent assets picked for downloading (but will not be downloaded as there is `--dry-run` param). If that cmd succeeds than verified copy of icloudpd ran for your account successfully and you can bisect to the problem.

skumka:
```
root@docker:/tmp/temp/data# docker run --rm -it icloudpd/icloudpd:1.17.3 icloudpd -u xxx@xx.xxx --recent 2 --dry-run -d ./
Unable to find image 'icloudpd/icloudpd:1.17.3' locally
1.17.3: Pulling from icloudpd/icloudpd
c926b61bad3b: Pull complete 
f821361375b1: Pull complete 
e619753a56d7: Pull complete 
eb0652d33051: Pull complete 
4f4fb700ef54: Pull complete 
2647709a09f9: Pull complete 
Digest: sha256:c28e9ac5bae49122bd8007be7d89582e793434341c539b7d1244f2e3bcb52898
[...]

Issue #793, opened 2024-02-21
Downloads consistently stall when including videos


dbyler:
## Overview
I've had successful photos-only downloads running icloud_photos_downloader via Docker on my Mac, but every time I try to include videos, it stalls within a couple minutes.

This command works for me:

`docker run -it --rm --name icloudpd -v /Volumes/MyDrive/Photos/data:/data -e TZ=America/Los_Angeles icloudpd/icloudpd:latest icloudpd --directory /data --username xxxxx@xxxxx.com --watch-with-interval 3600 --skip-videos --auto-delete`

However, when I remove the `--skip-videos` flag, the download invariably stalls after a couple minutes. 

## Steps to Reproduce

1. Open Docker Desktop for Mac (I'm running Docker 4.27.2 on macOS 14.1 on a M1 machine)
2. Run the shell command `docker run -it --rm --name icloudpd -v /Volumes/MyDrive/Photos/data:/data -e TZ=America/Los_Angeles icloudpd/icloudpd:latest icloudpd --directory /data --username xxxxx@xxxxx.com --watch-with-interval 3600 --auto-delete`
3. Watch and wait

## Expected Behavior

The full download executes successfully

## Actual Behavior

`icloudpd` usually manages to fetch a couple videos before stalling.

You can see three separate attempts in this screenshot, with Little Snitch showing the dropouts. The downloads don't restart.

![2024-02-21 13-31-51 Little Snitch Network Monitor -@2x](https://github.com/icloud-photos-downloader/icloud_photos_downloader/assets/57901/3657c5e0-e12c-4b58-9d76-096eebcf7351)

tsetsko:
I have the same issue - seems to cannot debug it and there is no pattern other than it stops at videos but it is at random intervals.

AndreyNikiforov:
Does the size of the video matter? I wonder if large videos are more prone to stalling... Just guessing at this point as I never experienced such behavior and cannot repro.

tsetsko:
@AndreyNikiforov I will be able to provide more information later today

dbyler:
FWIW I'm able to use the [compiled binary](https://github.com/icloud-photos-downloader/icloud_photos_downloader/releases) successfully with videos. No idea why Docker wasn't working but that was an effective workaround for me.

tsetsko:
@AndreyNikiforov I guess this might help:
[...]

PR #1114, opened 2025-04-07
Setting timeout for HTTP: requests does not hanging forever if the download speed is zero (addressing an issue when Apple does not response to HTTP GET)

naszta:
Addressing issue #787 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: timeout set to 30 seconds for HTTP requests [#793](https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/793)
+
 ## 1.27.4 (2025-04-15)
 
 - fix: broken pypi publishing [#1105](https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/1105)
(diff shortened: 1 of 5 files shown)


AndreyNikiforov:
seems that symptoms this PR fixes better match #793 WDYT?

naszta:
Yes, it might be. If we decide to merge: I would add it as config parameter for the session (so I would extend the commit a bit).

AndreyNikiforov:
> Yes, it might be. If we decide to merge: I would add it as config parameter for the session (so I would extend the commit a bit).

Oh, you mean so it applies to all requests to Apple? Sounds good. Can you think of any tests? Validating that timing out request is properly handled seems nice, but probably unrealistic to check all places. May be you can figure out how to test for original symptom (stuck requests)?

naszta:
Taking a look over the weekend.

naszta:
I added a similar test we have for `test_handle_session_error_during_photo_iteration_name_id7`

naszta:
Are we cool?

AndreyNikiforov:
> Are we cool?

Update changelog pls. The code seems fine.

naszta:
Done.

PR #1131, opened 2025-05-01
fix: add pagination support for album folder listing

asherawelan:
fix: add pagination support for album folder listing

Some iCloud accounts have more albums than are returned in a single response. Previously, the `_fetch_folders` method only fetched the first page of albums, ignoring any `continuationMarker` provided by the Apple Photos API.

This patch adds support for pagination, looping through all available pages and aggregating the results. This resolves issues where certain albums (e.g. "Harry Potter") were visible in the iCloud Web UI but missing from tools like `icloudpd --list-albums`.

Tested manually by confirming previously missing albums now appear.


asherawelan:
Tests and change log entry added

asherawelan:
@AndreyNikiforov Could you please re-review

PR #1152, opened 2025-06-05
feat: support conf User-Agent by environment

JonasGao:


AndreyNikiforov:
Why would anyone need to change user agent?

JonasGao:
Because apple respond 503 when i using icloudpd. I guess it is possible to bypass the traffic limit of apple by changing the user-agent. I haven't tested it yet, i will test later.

Issue #768, opened 2024-01-07
Support loading parameters from environment variables


leipert:
## Summary

It seems like `click` [supports loading parameters from env variables](https://click.palletsprojects.com/en/5.x/options/#values-from-environment-variables). This would make it easier to:

1. Configure a docker container in docker-compose with env variables
2. Potentially be a little more secure, because providing `--password` currently prints the password in certain contexts (e.g. `ps`)

## Context

Found my password exposed in the `ps` tree in the portainer interface.

<img width="1125" alt="Screenshot 2024-01-07 at 12 22 59" src="https://github.com/icloud-photos-downloader/icloud_photos_downloader/assets/2906107/5a3f9eb3-01d5-40f5-ab66-da6a8167d720">

choval:
You could pipe the password to icloudpd.

I have a shell script that I set as the entrypoint for the docker image.

```
#!/bin/sh
echo $PASSWD | /app/icloudpd_ex icloudpd ...
```

But I agree, if we could use an environment variable for it, it would be much better.

PR #1158, opened 2025-06-08
Add support for environment variable config (#768)

janw:
This adds support to configure iCloud PD using environment variables. 

The major benefit here is an easier configuration in e.g. containerized environments, where environment variables are much easier to handle and prevent the exposure of sensitive information through utilities such as `ps` or `top` from other users on the same system.

Caveat: to fully support env configuration, the click option `no_args_is_help` has been disabled. Otherwise, the user would have to specify at least one argument on the command line to allow every other variable to be taken from the environment. 

Happy to discuss this approach further. For example help could be shown if a certain set of required variables (username, directory, …) are not found in either env or args to ease user onboarding that way.

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -301,10 +301,13 @@ def report_version(ctx: click.Context, _param: click.Parameter, value: bool) -> 
 
 # Must import the constants object so that we can mock values in tests.
 
-CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"]}
+CONTEXT_SETTINGS = {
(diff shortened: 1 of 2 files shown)


AndreyNikiforov:
Can you show a comparison of the use in containerized environments of both approaches, with params and with env vars, please?

Using env var for passing sensitive information to containers is not recommended. Mounting files is a better way, e.g. see [doc](https://docs.docker.com/build/building/secrets/). If we are to introduce env vars, we may be better to block sensitive info to be passed through them.

PR #1207, opened 2025-08-08
feat: add snap packaging

jnsgruk:
Hi there

Opening this to gauge interest in keeping the [snap package](https://snapcraft.io/icloudpd) I maintain alongside the upstream code. My project for the snap is at https://github.com/jnsgruk/icloudpd-snap.

I've got some automation that automatically bumps the snap version each time a new release of this project is cut, so there hasn't been much of a lag. I've been running it for a few months this way without issue.

The snap adoption is almost at 300 active installs, and has grown steadily. This provides an instant way to install `icloudpd` on any Ubuntu machine with a single `snap install icloudpd`, so should lower the barrier nicely for some folks:

image

I've set things up in CI so the snap is released to the `candidate` channel - which means when you release anyone tracking that channel will get automatically upgraded to the latest version. There is then an additional action that is manually dispatched for "promoting" the snap to the `stable` channel. We can of course release straight to stable if your prefer, but that's relatively uncommon in the Snap ecosystem.

I'm happy to keep maintaining separately, or merge the project here. If it does get merged, I'm also happy to help troubleshoot any snap issues and keep it fresh as things evolve in that ecosystem - but I thought this was worth proposing as an option!

Thanks for all your work on this project! I'll be afk for a couple of weeks from today, but can try to answer any questions along the way - there is no rush from my side!

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -301,6 +301,37 @@ jobs:
         path: |
           dist/icloud*
   
+  build_linux_snap:
+    name: Build snap
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-24.04
+          - os: ubuntu-24.04-arm
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup LXD
+        uses: canonical/setup-lxd@v0.1.3
+
+      - name: Setup Snapcraft
+        run: sudo snap install --classic snapcraft
+
+      - name: Build snap
+        run: snapcraft --verbose
+      
+      - name: Upload artifacts 
+        uses: actions/upload-artifact@v4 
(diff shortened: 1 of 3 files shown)


AndreyNikiforov:
@jnsgruk What systems come with snap preinstalled, but without python v2+ and pip preinstalled?

I am open to exploring an option of adding new distribution mechanism to icloudpd project. I am not familiar with snap and need to read about it before accepting/merging PR - main focus is a balance between adding/maintain snap support vs additional user base that receives advantage of snap.

jnsgruk:
So this benefit of this method is that it's one command to install a single package on any Ubuntu system from 16.04 onwards - and any other machine where snapd happens to be installed.

Ubuntu doesn't ship pip out of the box as far as I know - user would first have to `apt install` it.

Adding a snap will also make it searchable in the Ubuntu software centre.

No rush like I say, and I totally understand you wanting to do some research! Feel free to ask any questions here if things are unclear.

PR #1291, opened 2025-12-16
Add /health endpoint for monitoring authentication and download state…

JonasGao:
Add a health endpoint for docker container to show the user loged apple in or not


stavros-k:
That would be great way to notify users instead of smtp. 
ie I have already a uptime-kuma monitoring health of my containers and send me alerts if something is unhealthy

Awesome

Issue #1299, opened 2026-01-02
ID7 suffix missing from adjusted/alternative filenames in v1.32.1+


spmp:
## Overview
ID7 suffix missing from adjusted/alternative filenames in v1.32.1+ 

A regression was introduced in v1.32.1 where the name-id7 file match policy is not being applied to adjusted and alternative size downloads. This causes filenames to be missing their 7-character ID suffix. Expected behavior (v1.32.0):

IMG_9815_QVRmWmV.HEIC              # original - has ID7 ✓
IMG_9815_QVRmWmV-adjusted.HEIC     # adjusted - has ID7 ✓
IMG_9815_QVRmWmV-medium.JPG        # medium - has ID7 ✓
Actual behavior (v1.32.1+):

IMG_9815_QVRmWmV.HEIC              # original - has ID7 ✓
IMG_9815-adjusted.HEIC             # adjusted - MISSING ID7 ✗
IMG_9815_QVRmWmV-medium.JPG        # medium - has ID7 ✓

Root cause: The refactoring in commit f52826c ("refactor: make AssetVersion immutable by moving filename disambiguation to caller") changed disambiguate_filenames() to call photo_asset.calculate_version_filename(), which uses self.filename - the raw filename WITHOUT file_match_policy applied. This causes filename comparisons and override generation to work with filenames missing the ID7 suffix. While commit f42639c introduced the filename_builder factory pattern to address this issue in base.py, the disambiguate_filenames() function was not updated to use it. Impact: Users with --file-match-policy name-id7 downloading adjusted or alternative sizes will get incorrectly named files, potentially causing:
- File organization issues
- Duplicate detection failures
- Confusion when the same photo has inconsistent naming across sizes

## Steps to Reproduce
1. icloudpd --recent 11 --directory /tmp/icloudpd-photos-adjusted-filename-id7-regression --folder-structure none --set-exif-datetime --keep-unicode-in-filenames --file-match-policy name-id7 --log-level debug --size original --size adjusted --size medium
2. Adjusted file names are wrong as compared to previous releases (and existing library) 

## Expected Behavior
Filenames as:
```
IMG_9815_QVRmWmV.HEIC              # original - has ID7 ✓
IMG_9815_QVRmWmV-adjusted.HEIC     # adjusted - has ID7 ✓
IMG_9815_QVRmWmV-medium.JPG        # medium - has ID7 ✓
```

## Actual Behavior
Filenames as:
```
[...]

PR #1300, opened 2026-01-02
fix: restore ID7 suffix for adjusted/alternative filenames with name-id7 policy

spmp:
This PR fixes a regression introduced in v1.32.1 where adjusted and alternative size downloads are missing their ID7 suffix when using the name-id7 file match policy.

Add optional base_filename parameter to disambiguate_filenames() in src/pyicloud_ipd/utils.py
Update call site in src/icloudpd/base.py to pass filename_builder(photo) as the base filename
Maintain backward compatibility with existing tests by making the parameter optional
Implementation Details: The fix follows the factory pattern introduced in commit f42639c by utilizing the filename_builder function that properly applies file_match_policy. When base_filename is provided (from production code), the function uses the static calculate_version_filename() with the correctly-built filename. When base_filename is None (from tests), it falls back to the photo_asset.calculate_version_filename() method for backward compatibility.

Fixes: #1299

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -660,7 +660,11 @@ def download_builder(
 
     try:
         versions, filename_overrides = disambiguate_filenames(
-            photo.versions_with_raw_policy(raw_policy), primary_sizes, photo, lp_filename_generator
+            photo.versions_with_raw_policy(raw_policy),
+            primary_sizes,
+            photo,
+            lp_filename_generator,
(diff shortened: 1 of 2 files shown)


spmp:
I checked issues and not PR's for this, there is another PR already out there to solve the same problem: https://github.com/icloud-photos-downloader/icloud_photos_downloader/pull/1250 - They should be the same, Claude likely wrote both of them 8)

Issue #1296, opened 2025-12-30
Potential PR's: Save Favorite as selectable rating in Exif AND a plugin system with example Immich integration


spmp:
Kia ora/Good day Team icloudpd,
I have been having some fun in your codebase on a few changes and want to discuss here before throwing PR's at you out of the blue. Apologies for my write first and PR later, I was excited and Claude code just fuelled that fire.

I have two changes - and maybe a bug fix

## Metadata management
(From PR #1301 )
"As a iPhotos user I often favorite my photos days after I take them, yet sync my photos daily so need to be able to update the favorite status post downloading. Favoriting is an important piece of metadata and should be included in files where possible for completeness"

"As a user I make changes to both EXIF and XMP metadata and need to ensure that syncing Favorites and other metadata from icloudpd will not overwrite the metadata changes I have made"
Add user facing CLI options:

* ` --favorite-to-rating`: Specify how `isFavorite` is expressed in the EXIF and XMP rating tag,
* `--process-existing-favorites` : Whether to process favorites on existing downloads, best if used in combination with --until-found using a sensible window which accounts for the number of photos between taking and rating
* `--metadata-overwrite`: Whether to overwrite existing metadata or not. Defaults to not overwriting metadata. For example if I use `--favorite-to-rating 1` and update the rating to 3 (I really like the photo), then a following run with `--process-existing-favorites` will not overwrite it. Wheras if the rating is unset then it is safe to set.

I have combined the `xmp_sidecar.py` and `exif_datetime.py` functionality (and tests) into metadata_management.py (and associated test_metadata_management.py to keep the metadata concerns together (XMP and EXIF are just i/o implementations of the same thing essentially), and cleaned up some of the calling logic in base.py, moving it into metadata_management.py to keep the concerns more distinct.

## Plugin framework
See PR: #1302 
"As an Ai enabled power user I want to be able to write custom hooks at various stages during the download process, with the context available to `icloudpd` to support various functions such as adding assets to my favorite media management system."

"As an Ai enabled power user and open source believer I want my changes to be available to the world to use and to be able to maintain my code as a non invasive addition to `icloudpd` or as a standalone project`

I have added a plugin system to `icloudpd` with hooks at various points in `basy.py`, together with a fully working Immich integration plugin. 
[...]

PR #1301, opened 2026-01-02
Feature/metadata management, rating to favorite, favorite updates, and metadata merge logic

spmp:
## Feature/metadata management, rating to favorite, favorite updates, and metadata merge logic

"As a iPhotos user I often favorite my photos days after I take them, yet sync my photos daily so need to be able to update the favorite status post downloading. Favoriting is an important piece of metadata and should be included in files where possible for completeness"

"As a <insert photo suite> user I make changes to both EXIF and XMP metadata and need to ensure that syncing Favorites and other metadata from icloudpd will not overwrite the metadata changes I have made"

In this PR I 

Add user facing CLI options:
* `--favorite-to-rating`: Specify how `isFavorite` is expressed in the EXIF and XMP `rating` tag,
* `--process-existing-favorites` : Whether to process favorites on existing downloads, best if used in combination with `--until-found` using a sensible window which accounts for the number of photos between taking and rating
* `--metadata-overwrite`: Whether to overwrite existing metadata or not. Defaults to not overwriting metadata. For example if I use `--favorite-to-rating 1` and update the rating to 3 (I really like the photo), then a following run with `--process-existing-favorites` will not overwrite it. Wheras if the rating is unset then it is safe to set.

I have combined the `xmp_sidecar.py` and `exif_datetime.py` functionality (and tests) into `metadata_management.py` (and associated `test_metadata_management.py`  to keep the metadata concerns together (XMP and EXIF are just i/o implementations of the same thing essentially), and cleaned up some of the calling logic in `base.py`, moving it into `metadata_management.py` to keep the concerns more distinct. 

I have formatted, linted, and ensured high test , whilst reducing some complexity in `base.py`, I hope you enjoy 🤗

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -763,36 +763,67 @@ def download_builder(
 
                 match download_result:
                     case DownloadMediaSuccess():
+                        logger.info("Downloaded %s", truncated_path)
+
+                        # Sync metadata to EXIF and XMP Sidecar for newly downloaded files
+                        # Must happen BEFORE set_utime to avoid mtime being overwritten
+                        sync_exif_metadata(
+                            logger=logger,
+                            download_path=download_path,
+                            photo=photo,
+                            created_date=created_date,
+                            favorite_to_rating=favorite_to_rating,
+                            set_exif_datetime=set_exif_datetime,
+                            process_existing_favorites=False,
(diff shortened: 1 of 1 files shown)


spmp:
Reference issue: #1296

PR #1310, opened 2026-01-12
Fix 2fa email notification date header format

porjo:
Current date format used in the 2fa email notice header is not [RFC 5322](https://www.rfc-editor.org/rfc/rfc5322#section-3.3) compliant and, in the case of Thunderbird, causes month and day to be swapped when sorting email by date e.g. 5th Jan becomes 1st of May

### DIFF
--- a/src/icloudpd/email_notifications.py
+++ b/src/icloudpd/email_notifications.py
@@ -37,7 +37,7 @@ def send_2sa_notification(
         smtp.login(smtp_email, smtp_password)
 
     subj = "icloud_photos_downloader: Two step authentication has expired"
-    date = datetime.datetime.now().strftime("%d/%m/%Y %H:%M")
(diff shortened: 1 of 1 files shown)


victorwon2001:
Thanks for addressing this. There is still one RFC 5322 issue here: `datetime.datetime.now()` is naive, so `%z` expands to an empty string. For example, the proposed expression produces a value like:

```text
Fri, 28 Aug 2026 15:36:18 
```

That leaves the required zone missing. One portable option on the supported Python versions is:

```python
from email.utils import format_datetime

date = format_datetime(datetime.datetime.now(datetime.timezone.utc))
```

This produces a numeric `+0000` zone. Could the existing email notification assertions also be updated to cover the complete `Date` header? They currently expect the old `01/01/2018 00:00` value.

PR #1325, opened 2026-03-22
fix: handle Apple's new trustedPhoneNumbers location in auth response

### LINKED ISSUE #1322: It seems that I don't receive the two-factor authentication code anymore when icloudpd requests it.
Since few days ago, when icloudpd (built in Synology Nas) requested to re-login, my devices (iPhone, iPad, Mac) didn't receive two-factor authentication code anymore. So my icloudpd just stopped working because I can't login my Apple ID. Is it blocked by apple? What should I provide to debug? Please help, thanks.

## Overview
[TIP]:  # ( DO NOT include your login-information when pasting in program output! )
[NOTE]: # ( Give a BRIEF summary about your problem )

## Steps to Reproduce
[NOTE]: # ( Provide a simple set of steps to reproduce this bug. Version of the app, OS/System, and minimum command line to reproduce. )
1. 
2. 
3. 

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen )

## Actual Behavior
[NOTE]: # ( Tell us what actually happens )

## Context
[NOTE]: # ( Tell us why expected behvior is valuable )

[NOTE]: # ( Give us any additional information you may have. )

matzekl:
## Summary

- Apple moved `trustedPhoneNumbers` from `twoSV.phoneNumberVerification` to `twoSV.bridgeInitiateData.phoneNumberVerification` in the HTML response from `https://idmsa.apple.com/appleauth/auth`
- This causes `get_trusted_phone_numbers()` to return an empty list, hiding the SMS 2FA option entirely
- Combined with iOS 26 removing the manual "Get Verification Code" button from Settings, users have no way to authenticate
- The fix checks both the original and new location, falling back to `bridgeInitiateData` when the original path is empty

Fixes #1322

## How I found this

Captured the actual Apple auth response and compared the JSON structure to what the parser expects:

```
# What the parser looks for (old):
twoSV.phoneNumberVerification.trustedPhoneNumbers → []  (empty)

# Where Apple now puts it (new):
twoSV.bridgeInitiateData.phoneNumberVerification.trustedPhoneNumbers → [{id: 1, ...}]
```

## Test plan

- [x] Verified the parser returns 0 devices with the unpatched code against a live Apple auth response
- [x] Verified the parser returns the correct trusted phone number after the fix
- [x] Built a patched Docker image and confirmed the SMS option (`a: (***) ***-**88`) now appears in the 2FA prompt
- [x] Backwards compatible: still checks the original location first

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pyicloud_ipd/sms.py
+++ b/src/pyicloud_ipd/sms.py
@@ -65,12 +65,17 @@ def parse_trusted_phone_numbers_payload(content: str) -> Sequence[TrustedDevice]
     parser = _SMSParser()
     parser.feed(content)
     parser.close()
+    twoSV = parser.sms_data.get("direct", {}).get("twoSV", {})
+    # Apple moved trustedPhoneNumbers into bridgeInitiateData.phoneNumberVerification (2026+)
     numbers: Sequence[Mapping[str, Any]] = (
-        parser.sms_data.get("direct", {})
-        .get("twoSV", {})
-        .get("phoneNumberVerification", {})
-        .get("trustedPhoneNumbers", [])
+        twoSV.get("phoneNumberVerification", {}).get("trustedPhoneNumbers", [])
     )
+    if not numbers:
+        numbers = (
+            twoSV.get("bridgeInitiateData", {})
+            .get("phoneNumberVerification", {})
+            .get("trustedPhoneNumbers", [])
+        )
     return list(item for item in map(_map_to_trusted_device, numbers) if item is not None)
 
 


billimek (approved):

dotnjet:
I run it with two accounts. One is connected to an iPhone with iOS 26.3 and the 2nd one with iOS 26.4. 
It is the update to iOS 26.4 causing it.

naszta:
Nice, can we merge it?

i-am-logger:
Just confirming this PR works end-to-end when combined with @iowk's `send_2fa_code()` change ([commit](https://github.com/iowk/icloud_photos_downloader/commit/4783098206a76eddc8b5f993398cc3e41b49e613)) referenced in #1322.

On its own, this PR restores the SMS device list in the prompt (no more empty `Please enter two-factor authentication code:`), but the trusted-device push from Apple still doesn't fire — `verify/trusteddevice/securitycode` needs an active PUT to trigger the push under iOS 26.4. With both patches applied locally I get the push notification on my iPhone immediately, can enter the code, and the session is saved as expected.

Could the maintainers consider pulling iowk's one-method addition into this PR (or merging both)? Together they form a complete iOS 26.4 fix.

boredazfcuk:
Doesn't 1335 fix fix both issues?

I rolled that one into my container and both SMS and Apple MFA seem to be working.

PR #1327, opened 2026-03-25
Fix: 2FA messages are not sent to trusted devices

myMartek:
This pull request includes the changes from #1325 but also adds another fix for the authentication flow.

It looks like Apple is using /auth/bridge/step/0 as another endpoint in order to actually send push messages with a 2FA code to all trusted devices. I added this request so that the 2FA flow will trigger code generation.

### DIFF
--- a/src/icloudpd/authentication.py
+++ b/src/icloudpd/authentication.py
@@ -152,6 +152,17 @@ def request_2sa(icloud: PyiCloudService, logger: logging.Logger) -> None:
 def request_2fa(icloud: PyiCloudService, logger: logging.Logger) -> None:
     """Request two-factor authentication."""
     devices = icloud.get_trusted_phone_numbers()
+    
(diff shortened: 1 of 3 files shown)


lathiat:
Trying out this fix, although the phone number list is retrieved and the push notification is sent to my devices, once I enter the code it fails showing an Incorrect Verification Code error.

It's not clear how much of the printed response I can include, but showing this part:
```
    "content": {
      "service_errors": [
        { 
          "code": "-21669",
          "title": "Incorrect Verification Code",
          "message": "Incorrect verification code.",
          "suppressDismissal": false
        }
      ],
      "hasError": true
    }
```

This alternative PR does work: https://github.com/icloud-photos-downloader/icloud_photos_downloader/pull/1335

eb3nezer:
I saw the same problem as @lathiat . I get the push notification, but when I enter the code it fails.

Issue #1305, opened 2026-01-06
Looking for new MAINTAINER for the project


AndreyNikiforov:
subj

AndreyNikiforov:
cc @menkej

menkej:
You did an amazing job @AndreyNikiforov ! Unfortunately I don't have capacity to dive in again. But I'm still following the development with half an eye.

So people: stand up and help!

tirrenian:
@AndreyNikiforov

Hope this helps:
https://news.ycombinator.com/item?id=46578921

Thanks for the app!

ndbroadbent:
Hi @AndreyNikiforov, thanks so much for all your work on this over the years! I'm very grateful that you were able to keep this project going for so long! It's amazing that people are still finding this useful after 10 years. I was surprised to see it on the front page of Hacker News again. Thanks for posting @tirrenian!

I've been very productive with AI agents lately so I might be able to help out again. I also run my own software business (https://docspring.com) so it would probably make a lot of sense for my company to help sponsor the maintenance and keep it working.

Please let me know how I can help!

These issues sound a bit concerning though:

* https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/1308
* https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/1306

It would be a shame if Apple shut it down by blocking/rate limiting requests from icloudpd.

tbrodbeck:
Hi @AndreyNikiforov and Hi everyone!

I have recently stumbled upon this tool dumping all my Photos from iCloud and found this tool a game-changer. Additionally, I saw the request for a new maintainer in the README and then found this issue.

I would be interest in maintaining or helping to maintain the project as well. In case you can give some more information about how much time you spent on maintenance or get some more specifics about the responsibilities as a maintainer, feel free to share. 

My background is a lot of Python Development for namely companies, mostly in the areas of AI, Data and DevOps.

AndreyNikiforov:
> These issues sound a bit concerning though

@ndbroadbent I suspect Apple's refusal to serve requests are for the calls from China mainland. I was unsuccessful in either obtaining VPN server in mainland for my own debugging or finding reliable partner there for troubleshooting.

chkuendig:
[...]

PR #1335, opened 2026-04-19
fix: restore 2FA for Apple's updated auth flow (2026+)

eukos:
Apple changed the 2FA flow around 2026. Two issues needed fixing:

1. Push trigger: the new flow requires a PUT to /verify/trusteddevice/securitycode (no body) to push the code to trusted devices. The old bridge/step/0 POST approach (PR #1327) triggers the code delivery but causes downstream failures.

2. Phone number parsing: Apple moved trustedPhoneNumbers into bridgeInitiateData.phoneNumberVerification in the auth HTML payload. Fall back to that path when the old path returns nothing.

### DIFF
--- a/src/icloudpd/authentication.py
+++ b/src/icloudpd/authentication.py
@@ -152,6 +152,16 @@ def request_2sa(icloud: PyiCloudService, logger: logging.Logger) -> None:
 def request_2fa(icloud: PyiCloudService, logger: logging.Logger) -> None:
     """Request two-factor authentication."""
     devices = icloud.get_trusted_phone_numbers()
+
+    # Trigger push notification to trusted devices before prompting for code.
+    # Apple's auth flow (2026+) requires a PUT to /verify/trusteddevice/securitycode
+    # to initiate code delivery. Failure is non-fatal — the user can still enter
+    # a code if it arrives via another path.
+    if not icloud.trigger_push_notification():
+        logger.debug("Failed to trigger 2FA push notification, continuing anyway")
+    else:
+        logger.debug("2FA push notification triggered")
+
     devices_count = len(devices)
     device_index_alphabet = "abcdefghijklmnopqrstuvwxyz"
     if devices_count > 0:
--- a/src/pyicloud_ipd/base.py
+++ b/src/pyicloud_ipd/base.py
@@ -694,6 +694,35 @@ class PyiCloudService:
 
         return parse_trusted_phone_numbers_response(response)
 
+    def trigger_push_notification(self) -> bool:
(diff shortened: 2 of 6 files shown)


lathiat:
Tested this PR and it works for me

nicoh88:
@AndreyNikiforov pls merge it. Thanks.

boredazfcuk:
Any chance this can be merged? I've pushed to my container and verified it's working for both Apple iOS popup notifications, plus SMS codes.

May I be added as a maintainer too? I won't be able to help much in regard to development, but hopefully will be able to keep the project alive with help from the community.

chkuendig:
@eukos FWIW, this only fixes the bug for the CLI flow, you'd also need to add your `src/icloudpd/authentication.py` changes to `request_2fa_web`.

@boredazfcuk maybe reach out in https://github.com/icloud-photos-downloader/icloud_photos_downloader/issues/1305 ?

boredazfcuk:
>FWIW, this only fixes the bug for the CLI flow, you'd also need to add your src/icloudpd/authentication.py changes to request_2fa_web.

@chkuendig Is the correct? I've applied this PR (and only this one) into latest version of my container and everything is working as expected from what I have found.

I primarily use SMS for re-authenticating, but tested the Apple popup code too.

Unless request_2fa_web is triggered during the process to store the password in the keyring. I remember it behaving differently some time ago?

chkuendig:
I use the WebUI MFA provider and it didn't work, which makes sense if you look at https://github.com/icloud-photos-downloader/icloud_photos_downloader/blob/3a97872f9f44ac49bda54b66b5170575bd22ff18/src/icloudpd/authentication.py#L107-L110

stevegore:
Can we do a release for this? Would be amazing. And a new container image.

boredazfcuk:
> Can we do a release for this? Would be amazing. And a new container image.

Just trying to figure it out lol.

I got the PR merged, added a test for it, merged a PR to fix the WebUI side of things, but when I built it on my test environment, there were issues with staticx due to the Python base image for Debian being bumped from Bookworm to Trixie. I also had to bump Alpine from 3.19 to 3.23 because of Cargo issues.

When it finally built all lovely in my test env, I triggered the release build and went to bed. I've had e-mail notifications overnight to say it's not been successful though.

It looks like Pypi and Docker images are all created automatically once the GitHub releases work.

Will be able to bump my own container as well after that, maybe even figure out how to trigger that automatically, but that's less important as my containers working at the minute as it's got this PR rolled in already.

boredazfcuk:
@stevegore 

Done

PR #1344, opened 2026-05-28
fix: add 2FA push notification trigger for WebUI (pr-1335 followup)

boredazfcuk:
Applies the same 2FA push notification fix from PR-1335 to the WebUI code path.

## Summary
Cherry-picks the fix from PR-1335 (CLI 2FA) to the WebUI implementation.

## Changes
- Modified `request_2fa_web()` in `src/icloudpd/authentication.py` to call `icloud.trigger_push_notification()` before prompting for the 2FA code
- This ensures Apple's updated auth flow (2026+) works correctly with WebUI

## Details
Apple's auth flow now requires a PUT to `/verify/trusteddevice/securitycode` to initiate code delivery. The fix adds this call before prompting the user, matching the CLI implementation.

## Testing
The WebUI uses the same underlying `PyiCloudService` methods, so it will work with the existing test infrastructure that was updated in PR-1335.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -66,9 +66,7 @@ RUN \
     python3 -m pip install --disable-pip-version-check --upgrade pip wheel && \
     python3 -m pip install --disable-pip-version-check -r requirements-pip.txt && \
     pip3 install --disable-pip-version-check . --group dev --group devlinux && \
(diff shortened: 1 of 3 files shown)


boredazfcuk (commented):
Failed.

PR #1367, opened 2026-08-06
Discover migrated Apple Photos primary zones

tural-ali:
## Summary

- query Apple Photos private zones before initializing the primary library
- accept active primary zone names such as `PrimarySync1` instead of assuming `PrimarySync`
- preserve the historical `--library PrimarySync` default for migrated accounts
- reuse discovered zone IDs when listing private libraries

## Problem

Some Apple accounts expose their primary Photos zone as `PrimarySync1` or another `PrimarySync*` value. Initializing every account with the hard-coded `PrimarySync` identifier causes Apple to return `ZONE_NOT_FOUND` even though authentication succeeded.

## Validation

- added regression coverage for `PrimarySync1` discovery
- added coverage for the missing-primary-zone error path
- built and started the patched package inside the current boredazfcuk/icloudpd image
- validated against a live account whose primary zone is `PrimarySync1`

The change keeps shared-library discovery unchanged and retains the existing CLI default.

### DIFF
--- a/src/icloudpd/base.py
+++ b/src/icloudpd/base.py
@@ -939,6 +939,12 @@ def core_single_run(
                         library_object: PhotoLibrary = icloud.photos.private_libraries[
                             user_config.library
                         ]
+                    elif user_config.library == "PrimarySync" and icloud.photos.zone_id.get(
+                        "zoneName", ""
+                    ).startswith("PrimarySync"):
+                        # Preserve the historical CLI default for accounts whose
+                        # primary zone has been migrated to names such as PrimarySync1.
+                        library_object = icloud.photos
                     elif user_config.library in icloud.photos.shared_libraries:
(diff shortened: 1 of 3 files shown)


victorwon2001 (commented):
CloudKit zones/list may return a mix of successful zone dictionaries and Zone Fetch Error dictionaries, and failed entries still contain zoneID. _fetch_zone_ids currently filters only deleted entries, so a failed PrimarySync* entry can be selected as the primary zone while a healthy entry later in the response is ignored. Please filter or explicitly handle entries containing reason/serverErrorCode before primary selection, and add a regression with a failed primary entry followed by a healthy one. Apple documents this response shape here: https://developer.apple.com/library/archive/documentation/DataManagement/Conceptual/CloudKitWebServicesReference/GettingAllZones.html. Also, ruff format --check reports that tests/test_photos_zone_discovery.py needs formatting.

PR #1371, opened 2026-08-18
Add resumable photo scan checkpoints

layneyu:
## What changed

Add an opt-in resumable scan checkpoint for photo albums.

When `ICLOUDPD_CHECKPOINT_DIR` is set, the photo scan stores the processed offset in an atomic JSON checkpoint. A restarted process restores the offset and continues from the saved position.

## Why

The downloader previously restarted photo metadata traversal from the newest assets after a transient iCloud/network failure. This caused expensive repeated scans for large libraries.

## Safety

Checkpointing is disabled by default. The checkpoint key includes the service endpoint, zone, query type, list type, and filter, so different album queries do not share offsets. The offset is advanced only after an asset is yielded, so an interrupted asset may be retried safely.

## Validation

- Added a unit test covering checkpoint creation, persistence, and restoration.
- `python3 -m pytest -q tests/test_photo_album_checkpoint.py` passed in a Python 3 container.
- `python3 -m py_compile src/pyicloud_ipd/services/photos.py` passed.

### DIFF
--- a/src/pyicloud_ipd/services/photos.py
+++ b/src/pyicloud_ipd/services/photos.py
@@ -1,7 +1,10 @@
 import base64
 import copy
+import hashlib
 import json
 import logging
+import os
+from pathlib import Path
 import re
 import typing
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


victorwon2001 (commented):
The checkpoint advances when the generator is resumed, before the caller knows whether the previous asset was handled successfully. In the main loop, a failed download returns False but iteration continues; requesting the next asset then persists an offset past the failed one. A restart therefore skips that asset instead of retrying it. Downstream filters have the same problem because skipped assets also advance the persistent offset. Could the checkpoint be committed only after the caller confirms the processing outcome, with account/filter identity included, and regression tests for failed downloads and changed filters? The current test only exercises manual increment/save/restore.

